PR #2, opened 2015-07-16
Finished ruby objects assessment

ghost:
Comfort: 5
Clarity: 5

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,6 +6,8 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 
 class Dog
+  attr_reader :name, :breed, :age
+
   def initialize name, breed, age
     @name, @breed, @age = name, breed, age
   end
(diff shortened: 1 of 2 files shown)


GA-MEB:
4.5 / 5 (-0.5 on Q3; missed that mistake earlier)

Excellent job, Arvin. Keep up the good work.

lib/assessment.rb
```diff
@@ -5,35 +5,45 @@
...
 # Question 3
 # Create a Person class, with attributes name, age, and location.
 # It should set all three attributes when you create a new Person.
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+class Person
+  attr_writer :name, :location
```
GA-MEB, line 27: Missed this mistake earlier... This should actually by `attr_accessor`, not `attr_writer`. `attr_writer` only gives you write access, while `attr_accessor` gives you both read _and_ write access. It's _extremely_ uncommon to use `attr_writer` - after all, how often should someone be able to change a value without being able to see what it is?

PR #3, opened 2015-07-16
Assessment Completed. All tests passing.

tiffboors29:
Comfort: 5
Clarity: 5

I used Google to research inheritance because I couldn't remember which way the < (or >)  was supposed to face. [Ruby Inheritance](http://rubylearning.com/satishtalim/ruby_inheritance.html)

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,35 +5,48 @@ Answer = OpenStruct.new
 # Create a class called Dog. It should have, as attributes, name, breed, and age.
 # Make all of those attributes read-only.
 
-##
-# your code here
-##
+class Dog
+  def initialize(name, breed, age)
+    @name = name
+    @breed = breed
+    @age = age
+  end
+end
 
 # Question 2
 # Create a new Dog object.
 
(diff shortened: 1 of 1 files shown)


GA-MEB:
4.5 / 5 (-0.5 on Q3; missed that mistake earlier)

Great work, Tiffany. I'm glad to see that you were able to Google your way out when you got stuck - it's good to get practice with that.

lib/assessment.rb
```diff
@@ -5,35 +5,48 @@
...
 # Question 3
 # Create a Person class, with attributes name, age, and location.
 # It should set all three attributes when you create a new Person.
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+class Person
+  attr_writer :name, :location
```
GA-MEB, line 27: Missed this mistake earlier - this should actually be `attr_accessor`, not `attr_writer`. `attr_writer` is almost never used because it only confers write access, _not read access_, and needing to use that particular combination is extremely uncommon.

PR #4, opened 2015-07-16
Completed objects assessment

ghost:


lib/assessment.rb
```diff
@@ -5,35 +5,61 @@
...
+    attr_reader :name
+    attr_reader :breed
+    attr_reader :age
+
+    def initialize(name, breed, age)
+        @name = name
+        @breed = breed
+        @age = age
+    end
+end
```
GA-MEB, line 19: Yep. Pro-tip:  All the `attr_...` methods can accept multiple symbols as arguments, so you can actually make all your `attr_reader` declarations on a single line.
e.g. `attr_reader :name, :breed, :age`.

GA-MEB:
4.5 / 5 (-0.5 on Q3 - sorry I missed it earlier)

Good job, Charlie. Would you say that you generally feel good about this stuff?

lib/assessment.rb
```diff
@@ -5,35 +5,61 @@
...
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+class Person
+
+    attr_writer :name
+    attr_reader :age
+    attr_writer :location
```
GA-MEB, line 35: Oops - missed this earlier. Although it's not as semantic, you should really be using `attr_accessor` for `:name` and `:location`, not `attr_writer`. The reason for this is that `attr_writer` creates a setter method _but not a getter method_. This is an extremely unusual pattern; usually, if we can write something, we can read it as well.

PR #5, opened 2015-07-16
Answers to assessment

ghost:
Comfort level: 5 out of 5

Clarity: Everything was clear until the rspec snafu.  If it weren't for that, I would say 5 out of 5.  Taking the snafu into account, I would say 3 out of 5.

Resources used: (1) My notes (2) Stack Overflow, but it wasn't helpful.  (It turned out that the issue was resolved when Jeff provided the revised rspec file.)

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,35 +5,48 @@ Answer = OpenStruct.new
 # Create a class called Dog. It should have, as attributes, name, breed, and age.
 # Make all of those attributes read-only.
 
-##
-# your code here
-##
+#
+class Dog
+  attr_reader :name, :breed, :age
+
+  def initialize(name, breed, age)
+    @name, @breed, @age = name, breed, age
+  end
+end
 
 # Question 2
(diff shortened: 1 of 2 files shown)


GA-MEB:
4.5 / 5 (-0.5 on Q3; sorry for missing it earlier)

Very nice work, Joe!

lib/assessment.rb
```diff
@@ -5,35 +5,48 @@
...
 # Create a Person class, with attributes name, age, and location.
 # It should set all three attributes when you create a new Person.
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+#
+class Person
+  attr_writer :name, :location
```
GA-MEB, line 29: Missed this earlier, but you should be using `attr_accessor` here, not `attr_writer`. `attr_writer` is almost never used, because it creates a 'setter' method _but not a 'getter one_ - in effect, it gives write access without giving read access.

PR #6, opened 2015-07-16
Completes assessment, adding changes made to assessment_spec.rb. Tests passing.

kristenlk:
Comfort: 4.5
Clarity: 5

I used Google to read up a bit on class inheritance and for some rspec clarification (specific sites: RubyMonk and Stack Overflow).

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -1,4 +1,5 @@
 require 'ostruct'
+require 'pry'
 Answer = OpenStruct.new
 
 # Question 1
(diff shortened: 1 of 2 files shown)


GA-MEB:
4.5 / 5 (-0.5 on Q3; sorry I missed this earlier)

Great job, Kristen.

lib/assessment.rb
```diff
@@ -20,20 +30,34 @@
 # Name and location should be writeable, but age should be read-only.
 
 ##
-# your code here
+class Person
+
+  attr_writer :name, :location
```
GA-MEB, line 35: Missed this earlier, but you should be using `attr_accessor` here, not `attr_writer`. `attr_writer` is almost never used because it gives us a 'write' method without giving us a 'read' one, and it's very unusual to be able to change something's value but not be able to see what it is.

PR #7, opened 2015-07-16
Completed assessment ruby objects, changes in specs for tests

lekova:
Comfort :5
Clarity :3

Had failing test on Person, and the message return got me confused. 
"Person has no setters and getters otherwise" 

I used bing search "ruby writable attribute"

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,13 +6,23 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 
 ##
-# your code here
+class Dog
+  attr_reader :name
+  attr_reader :breed
+  attr_reader :age
+
+  def initialize name, breed, age
+    @name = name
+    @breed = breed
+    @age = age
+  end
+end
 ##
 
 # Question 2
 # Create a new Dog object.
 
-Answer.dog = nil # replace nil with your code
+Answer.dog = Dog.new("Rex", "german shepherd", 2.5)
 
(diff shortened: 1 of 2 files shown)


lib/assessment.rb
```diff
@@ -6,13 +6,23 @@
...
+  attr_reader :name
+  attr_reader :breed
+  attr_reader :age
+
+  def initialize name, breed, age
+    @name = name
+    @breed = breed
+    @age = age
+  end
+end
```
GA-MEB, line 19: Yep, this is right. By the way, did you know that you can write all of those `attr_reader`s on a single line?
e.g.
`attr_reader :name, :breed, :age`

GA-MEB:
4.5 / 5 (-0.5 on Q3; sorry I missed this earlier)

Great job, Lina - keep up the good work.

lib/assessment.rb
```diff
@@ -20,20 +30,35 @@
 # Name and location should be writeable, but age should be read-only.
 
 ##
-# your code here
+class Person
+  attr_writer :name
+  attr_writer :location
```
GA-MEB, line 35: Missed this earlier, but these should really be `attr_accessor`, not `attr_writer`. `attr_writer` creates a 'setter' method but doesn't create a 'getter' one. Because it's very unusual for someone to be able to edit a value but not be able to see it, `attr_writer` is rarely used.

PR #8, opened 2015-07-16
Completed assessment

maksimhn:
Comfort: 4
Clarity: 3

the response syntax wasn't clear enough, I had to try few variants before it worked

I used this page for Classes inheritance clarification https://newcircle.com/bookshelf/ruby_tutorial/class_inheritance

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,13 +6,20 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 
 ##
-# your code here
+class Dog
+
+  attr_reader :name, :breed, :age
+
+  def initialize (name, breed, age)
+    @name, @breed, @age = name, breed, age
+  end
+end
 ##
 
 # Question 2
 # Create a new Dog object.
 
-Answer.dog = nil # replace nil with your code
+Answer.dog = Dog.new("lassie", "collie", 3)
 
 # Question 3
 # Create a Person class, with attributes name, age, and location.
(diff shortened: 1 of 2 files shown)


lib/assessment.rb
```diff
@@ -20,20 +27,44 @@
...
 # Question 5
 # Create another class called Developer that inherits from Person.
 # Give it a new public method called 'hire_for_job'
 
 ##
-# your code here
+class Developer < Person
+  def initialize (name, age, location)
+    @name, @age, @location = name, age, location
+  end
```
GA-MEB, line 56: This is actually not necessary - `Developer` inherits from `Person`, so it is able to run all of the methods that `Person` can (unless we specifically choose to overwrite them).

GA-MEB:
4.25 / 5 (-0.5 on Q3, -0.25 on Q5)

Very nice work, Max.

lib/assessment.rb
```diff
@@ -20,20 +27,44 @@
 # Name and location should be writeable, but age should be read-only.
 
 ##
-# your code here
+class Person
+
+  attr_writer :name, :location
```
GA-MEB, line 32: I missed this earlier, but you should be using `attr_accessor` there, not `attr_writer`. `attr_writer` only creates a 'setter' method, _not a 'getter' one_. Since it's really rare that you'd want write access to a variable but _not_ read access, `attr_writer` is used extremely rarely.

maksimhn, line 32: that’s what I though I should do at first ) then I asked Jeff and he recommended to stick with the specs :)

Maksim
dollydasheep@gmail.com

PR #9, opened 2015-07-16
completed

lparvinsmith:
made two commits, one before we were told to edit the spec file.

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,13 +6,20 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 
 ##
-# your code here
+class Dog
+  def initialize(name, breed, age)
+    @name, @breed, @age = name, breed, age
+  end
+
+  attr_reader :name, :breed, :age
+
+end
 ##
 
 # Question 2
 # Create a new Dog object.
 
-Answer.dog = nil # replace nil with your code
+Answer.dog = Dog.new("luca", "mutt", 2)
 
 # Question 3
 # Create a Person class, with attributes name, age, and location.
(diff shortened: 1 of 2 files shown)


lib/assessment.rb
```diff
@@ -20,20 +27,39 @@
...
+    @name, @age, @location = name, age, location
+  end
+
+  def set_name(name)
+    @name = name
+  end
+
+  def set_location(location)
+    @location = location
+  end
```
GA-MEB, line 41: Although these methods will allow you to change the values of `name` and `location`, we were actually looking to see you use `attr_accessor` here. `attr_accessor` will generate both 'getter' and 'setter' methods for both `@name` and `@location`.

lib/assessment.rb
```diff
@@ -20,20 +27,39 @@
...
 
 # Question 5
 # Create another class called Developer that inherits from Person.
 # Give it a new public method called 'hire_for_job'
 
 ##
-# your code here
+class Developer < Person
+  def hire_for_job
+    puts "you're hired, #{name}!"
```
GA-MEB, line 62: This will work if you're using `attr_reader` or `attr_accessor` on Person (or, alternatively, on Developer), because when you write `name`, it calls the newly generated 'getter' method `name`. However, if we hadn't included `attr_...`, this would not have worked; we need to refer to instance variables (such as 'name'), with an '@'; e.g. `@name`.

GA-MEB:
4.5 / 5 (-0.5 on Q3)

Nice job, Lara. Just watch out for that instance variable thing; it'll come up a lot.

PR #10, opened 2015-07-16
I am still having issues getting the syntax right and so I was unable…

GHolmes04:
Comfort: 2
Clarity: 4


GHolmes04:
Wanted to get it in before I ran out of time

I was unable to test my spec because my syntax was incorrect. This made it impossible to check my answers in time because I was spending all my time trying to get the right syntax

PR #11, opened 2015-07-16
Submits assessment

aemassa:
Not finished yet!

Comfort: 3
Clarity: 5

Used google to look into making a class inherit another class.  Still working.

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,35 +5,50 @@ Answer = OpenStruct.new
 # Create a class called Dog. It should have, as attributes, name, breed, and age.
 # Make all of those attributes read-only.
 
-##
-# your code here
-##
+class Dog
+  attr_reader :name, :breed, :age
+
+  def initialize(name, breed, age)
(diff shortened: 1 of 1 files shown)


lib/assessment.rb
```diff
@@ -5,35 +5,50 @@
...
 # Question 3
 # Create a Person class, with attributes name, age, and location.
 # It should set all three attributes when you create a new Person.
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+class Person
+  attr_writer :name, :location
```
GA-MEB, line 29: This should be `attr_accessor`, not `attr_writer`. `attr_writer` only creates a 'setter' method, _not a 'getter' method_, and this is such an unusual use case that `attr_writer` is rarely used.

GA-MEB:
4.5 / 5 (-0.5 on Q3)

Not finished? What are you talking about! You _nailed_ Q5. Nice work, Anna.

PR #12, opened 2015-07-16
Done

joebegley15:


lib/assessment.rb
```diff
@@ -5,35 +5,52 @@
...
+class Dog
+  attr_reader :name
+  attr_reader :breed
+  attr_reader :age
+  def initialize(name, breed, age)
+    @name = name
+    @breed = breed
+    @age = age
+  end
+end
```
GA-MEB, line 17: Yep, this is right! Pro-tip: you can actually write all those `attr_reader` calls in a single line, like this:
`attr_reader :name, :breed, :age`

lib/assessment.rb
```diff
@@ -5,35 +5,52 @@
...
 
 # Question 2
 # Create a new Dog object.
 
-Answer.dog = nil # replace nil with your code
+Answer.dog = Dog.new( "Max", "Lab", "9")
```
GA-MEB, line 22: Age should probably be a number, not a string. But I won't take any points off for it.

lib/assessment.rb
```diff
@@ -5,35 +5,52 @@
...
 # Create a Person class, with attributes name, age, and location.
 # It should set all three attributes when you create a new Person.
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+class Person
+  attr_accessor :name
+  attr_accessor :location
```
GA-MEB, line 31: Nailed it!!! You're one of few people who got this one perfectly. `attr_accessor` is the right tool here. Although `attr_writer` would give you a 'setter' method, it _won't_ give you a 'setter' method; since it's really unusual to want to be able to write to a variable but not be able to read from it, `attr_writer` is almost never used.

lib/assessment.rb
```diff
@@ -5,35 +5,52 @@
...
 
-##
-# your code here
-##
+class Developer
+#   Hires = []
+#   Developer << Person
+#   def hire_for_job
+#     Hires << Person
+end
```
GA-MEB, line 56: To make a class inherit from another class, you use the left carrot, `<`. 
e.g.
`class Developer < Person`

This gives `Developer` access to all of `Person`'s methods.

The `hire_for_job` method can be any method you like, as long as it's unique to `Developer` (i.e. `Person` _can't_ use it).

GA-MEB:
4 / 5 (-1 on Q5)

**Excellent** work, Joe. Keep up that effort - you're definitely making progress.

PR #13, opened 2015-07-16
Something's fucky... It isn't liking my attr_reader/attr_accessor met…

payne-chris-r:
…hods and I don't know why.

...I now know why. See next commit.

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,10 +6,12 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 class Dog
   def initialize(name,breed,age)
-    attr_reader :name
-    attr_reader :breed
-    attr_reader :age
+    @name, @breed, @age = name, breed, age
   end
+
+  attr_reader :name
+  attr_reader :breed
+  attr_reader :age
 end
 
 # Question 2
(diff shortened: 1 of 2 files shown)


lib/assessment.rb
```diff
@@ -4,36 +4,49 @@
...
+  def initialize(name,age,location)
+    #accessor allows for read and write, is there a write only? What is it? attr_writer
+    @name, @age, @location = name, age, location
+  end
 
-##
-# your code here
-##
+  attr_writer :name
+  attr_writer :location
```
GA-MEB, line 33: It kills me to tell you this, Chris, but `attr_accessor` is really the one to use. Why? Because why would we ever want to be able to write to a variable but _not_ read from it? It's such a bizarre use case that `attr_writer` is almost never used.

GA-MEB:
4.5 / 5 (-0.5 on Q3)

Great job on this, Chris. It's clear that you're building a strong understanding of this stuff.

lib/assessment.rb
payne-chris-r, line 33: 100% agree Matt, but I was just following the spec... I originally had
attr_accessor and it flagged it as an error!

On Thu, Jul 16, 2015 at 10:07 PM, Matt Brendzel notifications@github.com
wrote:

> In lib/assessment.rb
> https://github.com/ga-wdi-boston/ruby_objects_assessment/pull/13#discussion_r34857170
> :
> 
> > -##
> > -# your code here
> > -##
> > -  attr_writer :name
> > -  attr_writer :location
> 
> It kills me to tell you this, Chris, but attr_accessor is really the one
> to use. Why? Because why would we ever want to be able to write to a
> variable but _not_ read from it? It's such a bizarre use case that
> attr_writer is almost never used.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ga-wdi-boston/ruby_objects_assessment/pull/13/files#r34857170
> .

## 

Chris R. Payne
Email: 06cpayne@gmail.com

PR #14, opened 2015-07-16
Assessment completed

caraclarke:
Comfort: 2
Clarity: 3

Looked up objects and attr's.
http://www-rohan.sdsu.edu/doc/ruby/chp_04/classes.html
http://www.java2s.com/Code/Ruby/Class/CreatingReadableandWritableAttributesuseattraccessor.htm

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -5,35 +5,52 @@ Answer = OpenStruct.new
 # Create a class called Dog. It should have, as attributes, name, breed, and age.
 # Make all of those attributes read-only.
 
-##
-# your code here
-##
+class Dog
+  attr_reader :name, :breed, :age
+
+  def initialize(name, breed, age)
+    @name = name
+    @breed = breed
+    @age = age
+  end
+end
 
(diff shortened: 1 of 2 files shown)


lib/assessment.rb
```diff
@@ -5,35 +5,52 @@
...
 # Question 3
 # Create a Person class, with attributes name, age, and location.
 # It should set all three attributes when you create a new Person.
 # Name and location should be writeable, but age should be read-only.
 
-##
-# your code here
-##
+class Person
+  attr_writer :name, :location
```
GA-MEB, line 29: Although this will make `name` and `location` writeable, it won't make them readable. This is because `attr_writer` only creates 'getter' methods, _not 'setter' ones_. That's a really bizarre, unusual use case, so `attr_writer` is almost never used; instead, people use `attr_accessor`. That's really the right tool here too.

GA-MEB:
4.5 / 5 (-0.5 on Q3)

Great job, Cara!

PR #16, opened 2015-07-16
Finished Edition 

GHolmes04:
Comfort: 4
Clarity: 5

2nd update. Correct syntax errors and completed all tests 

Mostly used the notes and README for help. Also glanced at the Ruby doc to try to correct syntax errors. 

You can look at my original submission my looking at the previous commit.

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,13 +6,25 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 
 ##
-# your code here
-##
+class Dog
+  def initialize(name, breed, age)
+    @name = name
+    @breed = breed
+    @age = age
+  end
+attt_reader :name
+attt_reader :breed
+attt_reader :age
+end
 
 # Question 2
 # Create a new Dog object.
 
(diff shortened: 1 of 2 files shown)


GA-MEB:
3.5 / 5 on original (-0.5 on Q3, Q4, and Q5) but 4.5 / 5 (-0.5 on Q3) on the resubmit. Nice work, Greg!

lib/assessment.rb
```diff
@@ -20,20 +27,36 @@
...
+class Person
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
+  attr_reader :age
+  attr_writer :name
+  attr_writer :location
```
GA-MEB, line 39: Missed this earlier, but you should be using `attr_accessor` here, not `attr_writer`. The reason is that `attr_writer` only creates a 'setter' method, _not a 'getter' one_. Since wanting to be able to write to a property but not read from it is _extremely_ unusual, `attr_writer` is rarely used.

PR #31, opened 2015-10-23
assessment response

faetea:
The assessment questions felt straightforward, but the install process was a shitshow

### DIFF
--- a/lib/assessment.rb
+++ b/lib/assessment.rb
@@ -6,13 +6,21 @@ Answer = OpenStruct.new
 # Make all of those attributes read-only.
 
 ##
-# your code here
+class Dog
+  attr_reader :name, :breed, :age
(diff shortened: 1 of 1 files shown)


jrhorn424:
Thanks for the feedback. We'll do better next time. Sorry! :(

lib/assessment.rb
```diff
@@ -20,20 +28,33 @@
 # Name and location should be writeable, but age should be read-only.
 
 ##
-# your code here
+class Person
+  attr_reader :age
+  attr_accessor :name, :location
```
jrhorn424, line 33: This should be write-only. Instead, it's readable and writeable. How would you change that?

jrhorn424:
4.75/5 good job!

PR #35, opened 2015-10-23
added answers for assessment.rb

bbarbersox:
was unclear on how to make an instance method public


jrhorn424:
Instance methods are public by default.

jrhorn424:
5/5 great job! :stars:

PR #40, opened 2015-10-23
answers to ruby objects assessment

christinemoses:


lib/assessment.rb
```diff
@@ -6,13 +6,25 @@
 # Make all of those attributes read-only.
 
 ##
-# your code here
+class Dog
+  attr_reader: name
+  attr_reader: breed
+  attr_reader: age
```
jrhorn424, line 12: You can do these on one line if you like.

lib/assessment.rb
```diff
@@ -20,20 +32,37 @@
 # Name and location should be writeable, but age should be read-only.
 
 ##
-# your code here
+class Person
+  attr_writer: name
```
jrhorn424, line 36: `name` is a local variable here. It should be a value of a special kind. Not a string, but a...

lib/assessment.rb
```diff
@@ -20,20 +32,37 @@
...
+
 ##
 
 # Question 4
 # Create a new Person object using parameters "Dave", 32, and "Ohio".
 # Then, set that Person's location to "Somerville"
 
-Answer.dave = nil # replace nil with your code
-Answer.dave.location = nil # replace nil with your code
+Answer.dave = ("Dave", "Ohio") # Age can't be set as it's read-only
```
jrhorn424, line 53: This does not create a new object.

lib/assessment.rb
```diff
@@ -20,20 +32,37 @@
...
 
 # Question 5
 # Create another class called Developer that inherits from Person.
 # Give it a new public method called 'hire_for_job'
 
 ##
-# your code here
+class Developer < Person
+  attr_accessor: 'hire_for_job'
+  def initialize('hire_for_job')
```
jrhorn424, line 63: What is `hire_for_job` on this line? It is not a method.

lib/assessment.rb
```diff
@@ -6,13 +6,25 @@
...
+end
+
+
 ##
 
 # Question 2
 # Create a new Dog object.
 
-Answer.dog = nil # replace nil with your code
+Answer.dog = ("Sparky", "pug", 3) # replace nil with your code
```
jrhorn424, line 27: This doesn't create a new `Dog`. How would you do that?

jrhorn424:
1.5/5

PR #57, opened 2016-02-16
add answers to diagnostic

kpie89:


lib/diagnostic.rb
```diff
@@ -8,7 +8,16 @@
...
-# your code here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
```
GA-MEB, line 19: Missing an `end` there.

lib/diagnostic.rb
```diff
@@ -8,7 +8,16 @@
 # and location should only be writable.
 
 ##
-# your code here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
```
GA-MEB, line 14: Yep! Also would have accepted 

``` ruby
attr_reader :name, :age
attr_writer :name, :location
```

lib/diagnostic.rb
```diff
@@ -27,7 +37,12 @@
 # which returns the string "think think think".
 
 ##
-# your code here
+class Developer < Person
+
+  def initialize(solve_problems)
+    @solve_problems = "think think think"
+  end
+
```
GA-MEB, line 45: Your inheritance is correct, but you didn't create an instance method correctly.

``` ruby
class Developer < Person
  def solve_problems
    "think think think"
  end
end
```

lib/diagnostic.rb
```diff
@@ -95,7 +117,9 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
+# Mixins are a great way to use inheritance because you can use specific methods on a class instead of
+# having to call an entire class's set of methods and traits. It also has the advantage of being
```
GA-MEB, line 121: Sort of. It does give you the option of making your methods more modular.
But the biggest advantage to using mixins is that allows for multiple inheritance, since you can load _multiple_ mixins, but can only inherit from a single class (though, admittedly, you also get access to everything that the parent class inherits from).

lib/diagnostic.rb
```diff
@@ -108,7 +132,8 @@ def eat_meat(food)
 # which returns the string "kick, move, punch"
 
 class ComboAttack
-  attr_reader :moves, :damage
+  attr_accessor :moves
+  attr_reader :damage
```
GA-MEB, line 136: Sorry, there's no need to modify this.

lib/diagnostic.rb
```diff
@@ -141,5 +166,9 @@ def multiplier
     else
       1
     end
+
+    def get_possible_moves
+      puts @moves
+    end
```
GA-MEB, line 172: Because it comes after `private`, this method would be private, which was not specified.
Additionally, it's an instance method, not a class method.

lib/diagnostic.rb
```diff
@@ -141,5 +166,9 @@ def multiplier
     else
       1
     end
+
```
GA-MEB, line 169: To make the method chaining work, `punch`, `kick`, and `move` must each return `self`.

GA-MEB:
4.75 / 7 (-0.25 on Q1, -0.5 on Q3, -0.5 on Q6, -1 on Q7)

Let me know if you want any additional feedback on this.

PR #58, opened 2016-02-16
initial commit

jenweber:
Comfort: 3
Clarity: 3

I used http://www.railstips.org/blog/archives/2009/05/11/class-and-instance-methods-in-ruby/ to remind myself of class method syntax.

I could not get rspec spec to work, so code is untested.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,7 +8,18 @@ Answer = OpenStruct.new
 # and location should only be writable.
 
 ##
-# your code here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
+
 ##
 
 # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -95,7 +118,7 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
+# inheritance can be messy to maintain if it is more than 1 level deep, whereas a relationship to a mixin is more clear. Your code is more readable and easier to change without breaking things.
```
GA-MEB, line 121: Good answer!

lib/diagnostic.rb
```diff
@@ -131,6 +154,10 @@ def kick
     @damage *= multiplier
   end
 
+  def self.get_possible_moves
+    "kick, move, punch"
+  end
```
GA-MEB, line 159: Class method is correct.

GA-MEB, line 159: The way to make the method chaining work is for `punch`, `kick`, and `move` to return `self`.That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB:
6.5 / 7 (-0.5 on Q7)

Nice work, Jen. Come see me during a break and we can look at why Rspec wasn't working.

PR #59, opened 2016-02-16
ruby objects diagnostic

apalmer0:


lib/diagnostic.rb
```diff
@@ -62,8 +80,13 @@ def initialize
...
 # your answers here
-Answer.housecat_noise = nil
-#
+Answer.housecat_noise = "I am a Animal, and I go 'meow'"
+# Housecat.new creates a new instance of housecat, which inherits from cat. Cat
+# inherits from Animal. Calling the "say_hello" method would look first to the
+# housecat class and then to the cat class, unsuccessfully trying to find that
+# method. Finally it would look to the Animal class, where the "say_hello" is
+# written, and return the associated string. Because @sound is an instance
+# variable of the Housecat class, it is not overwritten by nil.
```
GA-MEB, line 89: Your reasoning is good, but you're mistaken about how the method call ultimately works. 
Although the method was written on the Animal class, self's value is determined at _runtime_. Since the method was called on `HouseCat.new`, at runtime the instance we were working with was a HouseCat, so `self` evaluates to `HouseCat`. It's almost as if the `say_hello` method were written within the HouseCat class.

lib/diagnostic.rb
```diff
@@ -119,16 +152,27 @@ def punch
...
   def kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
+    self
+  end
+
+  def damage
+    @damage
+  end
```
GA-MEB, line 172: We already have a reader for `@damage`, so this isn't necessary.

lib/diagnostic.rb
```diff
@@ -119,16 +152,27 @@ def punch
...
     @moves << "move #{direction}"
+    self
   end
 
   def kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
+    self
+  end
```
GA-MEB, line 168: Correct.

spec/diagnostic_spec.rb
```diff
@@ -1,4 +1,4 @@
-require_relative '../lib/assessment.rb'
+require_relative '../lib/diagnostic.rb'
```
GA-MEB, line 1: Shoot! Thought I'd pushed that change. Thanks for catching the bug.

GA-MEB:
6.5 / 7 (-0.5 on Q4)

Nice work.

lib/diagnostic.rb
apalmer0, line 89: Thanks for taking the time to explain this, that's a lot clearer now. So it
should be "I am a housecat and I go 'meow'"?

GA-MEB, line 89: Yep, that's right.

PR #60, opened 2016-02-16
Ruby Tues objects diagnostic

msyao:


lib/diagnostic.rb
```diff
@@ -27,7 +39,11 @@
 # which returns the string "think think think".
 
 ##
-# your code here
+class Developer < Person
+  def solve_problems
+    returns "think think think"
```
GA-MEB, line 44: It's `return`, not `returns`. Also, an explicit `return` is not necessary in Ruby, since a Ruby method will default to returning the last evaluated result in its body.

lib/diagnostic.rb
```diff
@@ -95,7 +117,7 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
+# Mixin modules allows class methods to add functionality from multiple places and share functionality between classes without using inheritance.
```
GA-MEB, line 120: Yes, adding code from multiple sources is one of the chief advantages of composition over inheritance.

lib/diagnostic.rb
```diff
@@ -115,22 +137,26 @@ def initialize
...
+  def self.kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
   end
 
+  def get_possible_moves
+    return "kick, move, punch"
+  end
+
```
GA-MEB, line 159: Sorry, you got these mixed up. Writing `self.method_name` is how you define a class method - it should be `self.get_possible_moves`.

To get the method chaining to work, `punch`, `kick`, and `move` should return `self`. That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB:
5.5 / 7 (-0.5 on Q3, -1 on Q7)

Let me know if you'd like any additional feedback.

PR #62, opened 2016-02-16
Response to diagnostic.

dumplingpuff:


lib/diagnostic.rb
```diff
@@ -96,8 +122,10 @@ def eat_meat(food)
 
 ##
 # your answer, in comments, here
-#
-#
+# Using modules over direct inheritance allows your code to be more flexible
+# and easier to read. You can easily define certain methods and variables
+# and transition them over to multiple classes without having to worry
+# about other things getting mixed in.
```
GA-MEB, line 128: Your first point is true. Your second point is... meh. That's sorta true of both inheritance and composition. One example of an advantage that composition has over inheritance is that you can use an arbitrary number of modules as mixins, but can only inherit from one other class. And while the parent class also may inherit methods and properties from other classes, that can get really messy. Hence your first point.

lib/diagnostic.rb
```diff
@@ -107,9 +135,16 @@ def eat_meat(food)
...
 
+## Question 7 is not complete I was in the middle of analyzing it,
+# I did create a class method, but haven't fixed the ComboAttack class.
+
 class ComboAttack
   attr_reader :moves, :damage
 
+  def self.get_possible_moves
+    "kick, move, punch"
+  end
```
GA-MEB, line 146: Correct, this is how you create a class method.

GA-MEB, line 146: The way to make the method chaining work is for `punch`, `kick`, and `move` to return `self`.That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB:
6.25 / 7 (-0.25 on Q6, -0.5 on Q7)

Nice work.

PR #63, opened 2016-02-16
response to diagnostic

jdamiron:


lib/diagnostic.rb
```diff
@@ -8,7 +8,11 @@
 # and location should only be writable.
 
 ##
-# your code here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+end
```
GA-MEB, line 15: Halfway there. This doesn't satisfy the direction "All three attributes should be defined when you instantiate a new Person.". To do that, you'll need to define an `initialize` method.

lib/diagnostic.rb
```diff
@@ -17,8 +21,12 @@
...
 
 ##
-# your answers here
-Answer.dave = nil
+def initialize("Dave", 32, "Ohio")
+  @name = name
+  @age = age
+  @location = "Somerville"
+end
+Answer.dave = @location
```
GA-MEB, line 29: Sorry, this isn't right. We wanted you to instantiate a new Person. That means calling `.new`. 

``` ruby
dave = Person.new("Dave", 32, "Ohio")
dave.location = "Somerville"
Answer.dave = dave
```

lib/diagnostic.rb
```diff
@@ -27,7 +35,11 @@
 # which returns the string "think think think".
 
 ##
-# your code here
+class Developer < Person
+  def solve_problems
+  "think think think"
+  end
+end
```
GA-MEB, line 42: Yep.

lib/diagnostic.rb
```diff
@@ -62,8 +74,15 @@ def initialize
...
-Answer.housecat_noise = nil
-#
+Answer.housecat_noise = "I am a HouseCat, and I go meow"
+
+# This would be the output because the class HouseCat is inheritted from the
+# class Cat,which is inheritted from the class Animal, which has access to the
+# animal methods. The instance variable sound for HouseCat is equal to a string
+# that says meow. HouseCat inherits the say_hello method from the Animal class
+# and is able to evaluate that method using it's own declared instance variable
+# for sound.
```
GA-MEB, line 84: All correct, particularly the following

> HouseCat inherits the say_hello method from the Animal class and is able to evaluate that method using it's own declared instance variable for sound.

Since it's HouseCat calling the method, that's also why `self` refers to HouseCat, and not Animal.

lib/diagnostic.rb
```diff
@@ -95,8 +119,8 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
-#
+# Using mixins makes the code you write more semantic and easier to read
+# for future developers and future you.
```
GA-MEB, line 123: This is true. It's not one of the primary reasons why composition is preferred, though.
One big reason is the problem of multiple inheritance. A class can mix in an arbitrary number of modules. Classes can inherit from other classes, and gain methods and properties that way; however, this is less flexible than using mixins, and harder to organize.

GA-MEB:
5 / 7 (-1 on Q2, -1 on Q7)

Let me know if you want any additional feedback.

PR #64, opened 2016-02-16
ajordon diag response

ajordon:
When running the code for the last question I was getting a "undefined method move" for some reason. Not enough time to debug

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,7 +8,17 @@ Answer = OpenStruct.new
 # and location should only be writable.
 
 ##
-# your code here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 # Question 2
@@ -18,7 +28,9 @@ Answer = OpenStruct.new
 
 ##
 # your answers here
-Answer.dave = nil
+david = Person.new("Dave", 32, "Ohio")
+david.location = "Boston"
+Answer.dave = david
 #
 
 # Question 3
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -18,7 +28,9 @@
 
 ##
 # your answers here
-Answer.dave = nil
+david = Person.new("Dave", 32, "Ohio")
+david.location = "Boston"
```
GA-MEB, line 32: "Somerville", not "Boston", but you've got the idea.

lib/diagnostic.rb
```diff
@@ -95,7 +118,8 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
+# Composition yields beter looking code which is more organized and easier to
+# read.
```
GA-MEB, line 122: Sure, this is true. Why do you think that is?

lib/diagnostic.rb
```diff
@@ -116,19 +140,23 @@ def initialize
...
-    @moves << 'kick'
-    @damage += 10
-    @damage *= multiplier
+    self.moves << 'kick'
+    self.damage += 10
+    self.damage *= multiplier
+  end
+
+  def get_possible_moves
+    p "kick, move, punch"
```
GA-MEB, line 159: Nope, you've got this one kinda tuned around.

GA-MEB, line 159: To define a class method, use `self` in the method signature.
i.e.

``` ruby
def self.get_possible_moves
  #...
end
```

GA-MEB, line 159: The way to make the method chaining work is for `punch`, `kick`, and `move` to return `self`.That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB:
6 / 7 (-1 on Q7)

Nice work. Let me know if you'd like additional feedback.

PR #67, opened 2016-02-16
Response to diagnostic

seatuna:
Fist to five = 3.5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,7 +8,17 @@ Answer = OpenStruct.new
 # and location should only be writable.
 
 ##
-# your code here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 # Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -17,8 +27,9 @@
 # Finally, assign the modified Person to `Answer.dave` below.
 
 ##
-# your answers here
-Answer.dave = nil
+dave = Person.new('Dave', 32, 'Ohio')
+@location = 'Somerville'
```
GA-MEB, line 31: Not quite. Person has a 'setter' method for location (generated by `attr_writer` in Q1) called `location=`.
To set the property value, we can just write `dave.location = "Somerville"`.

lib/diagnostic.rb
```diff
@@ -135,11 +159,12 @@ def kick
...
+      @damage = 1.5
     when ['kick', 'punch', 'up']
-      2
+      @damage = 2
     else
-      1
+      @damage = 1
     end
   end
+
```
GA-MEB, line 169: No, sorry.

GA-MEB, line 169: The way to make the method chaining work is for `punch`, `kick`, and `move` to return `self`.That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB, line 169: To define a class method, use `self` as part of the method definition.
i.e.

``` ruby
class ComboAttack
  def self.get_possible_moves
    #... 
  end
end
```

GA-MEB:
5.5 / 7 (-0.5 on Q2, -1 on Q7)

Let me know if you have any questions or would like additional feedback.

PR #68, opened 2016-02-16
wrote answer

Knguyen21:


lib/diagnostic.rb
```diff
@@ -18,7 +29,8 @@
 
 ##
 # your answers here
-Answer.dave = nil
+Answer.dave =  Person.new("Dave", 32, "Ohio")
+Answer.dave.location = "Somerville"
```
GA-MEB, line 33: Sure, this works.

lib/diagnostic.rb
```diff
@@ -63,7 +80,10 @@ def initialize
 ##
 # your answers here
 Answer.housecat_noise = nil
-#
+I am a HouseCat, and I go 'meow'
```
GA-MEB, line 83: Should be a String, otherwise tests won't work for it.

lib/diagnostic.rb
```diff
@@ -63,7 +80,10 @@ def initialize
...
+
+HouseCat inherits from Cat class, which inherits from the Animal Class. The Animal Class has a
```
GA-MEB, line 85: This should be a comment.

lib/diagnostic.rb
```diff
@@ -63,7 +80,10 @@ def initialize
...
+say_hello method.
```
GA-MEB, line 86: Otherwise correct.

lib/diagnostic.rb
```diff
@@ -95,8 +120,10 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
-#
+
+Advantages of using composition over direct inheritance : Modules are more managable. Inheritance
```
GA-MEB, line 124: This should be a comment.

lib/diagnostic.rb
```diff
@@ -95,8 +120,10 @@ def eat_meat(food)
...
+lead to diamond shape inheritance. Modules are more flexible. Because include is a regular method
+call, Modules can be managed at runtime.
```
GA-MEB, line 126: Actually, Ruby doesn't have the 'diamond problem' because it only lets you inherit from one class. However, the rest of your answer is correct.

lib/diagnostic.rb
```diff
@@ -142,4 +170,8 @@ def multiplier
       1
     end
   end
+
+  def get_possible_moves
+    'kick, move, punch'
+  end
```
GA-MEB, line 176: This should be `self.get_possible_moves` - as written, it's an instance method, not a class method.

lib/diagnostic.rb
```diff
@@ -142,4 +170,8 @@ def multiplier
       1
     end
   end
+
```
GA-MEB, line 173: The way to make the method chaining work is for `punch`, `kick`, and `move` to return `self`.That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB:
6 / 7 (-0.25 on Q6, -0.75 on Q7)

Nice work. Let me know if you'd like any additional feedback on this.

PR #69, opened 2016-02-16
partially answered

byau:


lib/diagnostic.rb
```diff
@@ -62,8 +87,11 @@ def initialize
 
 ##
 # your answers here
-Answer.housecat_noise = nil
-#
+Answer.housecat_noise = "I am a HouseCat, and I go 'meow'"
+# While say_hello invoke the method in Animal, but HouseCat.new invoked creating
+# new object under the sub class HouseCat, which self.class.name will refer to
+# the class in which it was invoked, which is HouseCat.  @sound is unquie to
+# the subclass Housecat, so it turn "meow".
```
GA-MEB, line 94: This explanation is a little convoluted, but I think I get your meaning.

You are incorrect, though, that `@sound` is unique to HouseCat; Animal has a `@sound` property too.

lib/diagnostic.rb
```diff
@@ -96,8 +132,9 @@ def eat_meat(food)
 
 ##
 # your answer, in comments, here
-#
-#
+# Direct inheritance can only have 1 direct parent class (more if you think
+# grandparents), but you can include as many mixins into a class.
+# This is you more flexible to define specific features in a subclass
```
GA-MEB, line 137: Good answer.

lib/diagnostic.rb
```diff
@@ -131,6 +168,10 @@ def kick
     @damage *= multiplier
   end
 
+  def get_possible_moves
+    puts "kick, move, punch"
+  end
+
```
GA-MEB, line 174: The way to define a class method is to use `self` in the method name.
i.e.

``` ruby
def self.get_possible_moves
  puts "kick, move, punch"
end
```

GA-MEB, line 174: The way to make the method chaining work is for `punch`, `kick`, and `move` to return `self`.That way, `ComboAttack.new.punch.move('left').kick.damage` is equivalent to

``` ruby
combo = ComboAttack.new
combo.punch
combo.move('left')
combo.kick
combo.damage
```

GA-MEB:
6 / 7 (-0.25 on Q4, -0.75 on Q7)

Nice work. Let me know if you have any questions, or want additional feedback on this.

PR #76, opened 2016-04-19
Reply to diagnostic

cuprous:
Lion test case does not match prompt expectations (ROAR! != ROAR)
Test expects a dog class, but no questions bring up dogs.

Comfort: 5
Clarity: 5

Google, class notes, and my thoughts were used to complete this diagnostic.


RDegnen:
7 of 7. Thanks for pointing out the random dog test

PR #79, opened 2016-04-19
Ruby Objects Diagnostic

BenGitsCode:
Clarity 2
Comfort 1

I'm hitting a really flat wall with Ruby and I really want to spend some one on one
or group review time trying to understand even the tiniest bit more of what we've
gone over in the past week. Some of js stuck bits at a time and then found cohesion
through time and repetition but I feel like not even the smallest bits of Ruby are
sticking with me yet.


lib/diagnostic.rb
```diff
@@ -13,7 +13,16 @@
...
-# your code here
+class Person
+  attr_reader :age
+  attr_accessor :name
+  attr_writer :location
+  def initialize
+    @name = name
+    @age = age
+    @location = location
+  end
```
RDegnen, line 24: you forgot to pass in parameters here e.g., `def initialize (name, age, location)`

RDegnen:
1. Yea lets definitley schedule a one on one to go over ruby

PR #80, opened 2016-04-19
Not done =(

louarnos:


lib/diagnostic.rb
```diff
@@ -67,8 +84,10 @@ def initialize
 
 ##
 # your answers here
-Answer.housecat_noise = nil
-#
+Answer.housecat_noise = 'I am a HouseCat, and I go 'meow''
```
RDegnen, line 87: watch out with the single quotes instead of the double quotes. Double quotes allowing string interpolation but single quotes do not

RDegnen:
4 of 7

PR #81, opened 2016-04-19
Response to diagnostic

rdelvallej32:


lib/diagnostic.rb
```diff
@@ -124,16 +148,23 @@ def punch
...
 
   def kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
+    self
+  end
+
+  def get_possible_moves
+    "#{@kick}, #{@move}, #{@punch}"
```
RDegnen, line 167: just `"kick, move, punch"` because there are no kick move and punch instance variables

RDegnen:
6.5 of 7 fabulous!

PR #83, opened 2016-04-19
Complete ruby-objects-diagnostic

sjf125:


lib/diagnostic.rb
```diff
@@ -147,4 +171,7 @@ def multiplier
       1
     end
   end
+  def get_possible_moves
+    "kick, move, punch"
+  end
```
RDegnen, line 176:
``` rb
def self.get_possible_moves
    "kick, move, punch"
  end
```

RDegnen:
6 out of 7 nice!

PR #85, opened 2016-04-19
Response

philschoof:
3
3


RDegnen:
2 of 7. It seems you were a little confused about where the answers were supposed to go. Do you want to go over this?

philschoof:
Yeah I couldn't get the answers or tests to work in time. Do you have time
to meet today?

PR #107, opened 2016-06-15
Ruby objects diagnostic

RoobScoobs:
complete objects diagnostic

comfort: 2
clarity: 2

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -13,7 +13,15 @@ Answer = OpenStruct.new
 # and location should only be writable.
 
 ##
-# your code here
+class Person
+  attr_accesor :name
+  attr_reader :age
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location =  location
+  end
+end
 ##
 
 # Question 2
@@ -22,8 +30,9 @@ Answer = OpenStruct.new
 # Finally, assign the modified Person to `Answer.dave` below.
 
 ##
-# your answers here
-Answer.dave = nil
+new_person = Person.new('Dave', 32, 'Ohio')
+new_person[:location] = 'Somerville'
+Answer.dave = new_person
 #
 
 # Question 3
@@ -32,7 +41,11 @@ Answer.dave = nil
 # which returns the string "think think think".
 
 ##
-# your code here
+class Developer < Person
+  def solve_problems
+    p 'think think think'
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -13,7 +13,15 @@
...
-# your code here
+class Person
+  attr_accesor :name
+  attr_reader :age
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location =  location
+  end
+end
```
berziiii, line 24: Correct but don't forget to add Location to a attr. In the question, it asked for it to be writable but not readable. So it would be a `attr_writer`

lib/diagnostic.rb
```diff
@@ -22,8 +30,9 @@
 # Finally, assign the modified Person to `Answer.dave` below.
 
 ##
-# your answers here
-Answer.dave = nil
+new_person = Person.new('Dave', 32, 'Ohio')
+new_person[:location] = 'Somerville'
```
berziiii, line 34: Almost. You would do `new_person.location` since we would have `location` as a writable attr.

lib/diagnostic.rb
```diff
@@ -32,7 +41,11 @@
 # which returns the string "think think think".
 
 ##
-# your code here
+class Developer < Person
+  def solve_problems
+    p 'think think think'
+  end
+end
```
berziiii, line 48: Correct.

lib/diagnostic.rb
```diff
@@ -66,9 +79,17 @@ def initialize
 end
 
 ##
-# your answers here
-Answer.housecat_noise = nil
-#
+
+Answer.housecat_noise = 'I am a HouseCat, and I got meow.'
```
berziiii, line 83: Correct.

lib/diagnostic.rb
```diff
@@ -91,7 +112,12 @@ def eat_meat(food)
 end
 
 ##
-# your code here
+class Lion < Cat
+  include Carnivorous
+  def roar
+    p 'ROAR'
+  end
+end
```
berziiii, line 120: Correct.

lib/diagnostic.rb
```diff
@@ -100,9 +126,10 @@ def eat_meat(food)
 # Write your answer as a comment in the section below.
 
 ##
-# your answer, in comments, here
-#
-#
+# Mixins only exist to add functionality to a class and exist independently of the classes,
+# so the advantage here is that you can add them to different classes that are not related
+# to each other. Using direct inheritance will simply extend the class and it's dependent on
+# establishing some sort of relationship.
```
berziiii, line 132: Correct.

berziiii:
6 / 6. Nice work!

PR #109, opened 2016-08-08
Add answers to Ruby objects diagnostic

jlmitch1989:
Comfort:2
Clarity:3

I used the resources from class.  I felt like I had somewhat
of an understanding, but I'm not really sure.


RealWeeks:
I can't really grade this because the linter blows up.  Run `rspec rpec` to see your errors.  If you want me to go over this correct the syntax errors so the linter runs (I assume this will also fix a bunch of code as well).

PR #113, opened 2016-08-08
answered questions

cdelsoin:


RealWeeks:
Tests, won't run. Run `rspec spec` to run the test and correct the errors.  If you want I will grade once a pull request where test can be run is made.

PR #114, opened 2016-08-08
Completed ruby-objects-diagnostic

natGalynker:
-got to all the questions
-the last one stumped me because I could tell it needed to be fixed but the
directions were a bit confusing. I was not sure how to go about fixing it. I
would have created more than one class to handle the actions rather than having
so many methods in one class.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -24,8 +24,18 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
-##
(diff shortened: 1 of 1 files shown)


RealWeeks:
I cannot grade this as tests don't run.  Run `rspec spec` to run test and correct the errors.  If you would like I can grade this once the errors are fixed. Read you errors carefully, they will tell you exactly what you need to do.

PR #115, opened 2016-08-08
Answered questions, except for last one.

alex-braun:
Last question not completed.


RealWeeks:
I can't run test here to grade, because the test blow up.  If you want I can grade once you correct the code so test can process.  Run `rspec spec` to check tests.

PR #116, opened 2016-08-08
Add response

EchoXin:


RealWeeks:
Tests don't run, so I cannot grade.  If you would like, you can resubmit once tests can run.  Use `rspec spec` to run tests

PR #146, opened 2016-11-14
Diagnostic Mon Nov 14

RpRoderick:
Ruby Objects

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -24,7 +24,14 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+Class Person(name, age, location)
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+    @name = name
+    @age = age
+    @location = location
+end
 ##
 
 ##
@@ -43,7 +50,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+Class Developer < Person
+  def solve_problems
+    put 'think think think'
+  end
+end
 ##
 
 ##
@@ -74,16 +85,20 @@ end
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "'lick... lick...''I am a animal and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+Calling HouseCat, initilizes @sound as 'meow', goes up to Cat, puts 'lick...lick'
+which is an instance method of Class Animal, which calls say_hello, which puts
+"I am a #{self.class.name}, and I go '#{@sound}'", self refers to Class Animal, and @sound
+refers to an initialized sound, which was defined with HouseCat.
 ##
 # Define a new class, 'Lion', which (a) inherits from 'Cat',
 # (b) uses the 'Carnivorous' module below as a mixin, and
 # (c) adds a new method called `roar`, which prints out "ROAR!"
+
 
 module Carnivorous
   def can_eat_meat?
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -24,7 +24,14 @@
...
 ##
-# your response here
+Class Person(name, age, location)
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+    @name = name
+    @age = age
+    @location = location
+end
```
neugierige, line 34: Very close! The parameters with which you initialize instances of a class belong in the `initialize` method.

lib/diagnostic.rb
```diff
@@ -24,7 +24,14 @@
...
 ##
 
 ##
```
neugierige, line 37:
```
dave.location = 'Somerville'
Response.dave = dave
```

lib/diagnostic.rb
```diff
@@ -43,7 +50,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+Class Developer < Person
+  def solve_problems
+    put 'think think think'
```
neugierige, line 55: 2 things: 1) in Ruby, it's `puts` not `put`; 2) we want the string to be _returned_, which, if you'll remember, if fundamentally different from printing to the console

lib/diagnostic.rb
```diff
@@ -74,17 +85,21 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "'lick... lick...''I am a animal and I go 'meow'"
```
neugierige, line 88: Solution: `Response.housecat_noise = "I am a HouseCat, and I go 'meow'"`

`HouseCat` inherits from `Cat`, which inherits from `Animal`. The `'lick... lick...'` string results from the method `groom` which is not called by `HouseCat.new.say_hello`.

lib/diagnostic.rb
```diff
@@ -74,17 +85,21 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+Calling HouseCat, initilizes @sound as 'meow', goes up to Cat, puts 'lick...lick'
+which is an instance method of Class Animal, which calls say_hello, which puts
+"I am a #{self.class.name}, and I go '#{@sound}'", self refers to Class Animal, and @sound
+refers to an initialized sound, which was defined with HouseCat.
```
neugierige, line 96: remember that whenever an object's properties or methods are called, it first looks at it's own class, and only when it doesn't find it, does it traverse up the inheritance tree

lib/diagnostic.rb
```diff
@@ -100,15 +115,18 @@ def eat_meat(food)
 end
 
 ##
-# your response here
-##
+Class Lion < Cat
+  def roar
+    puts 'ROAR!'
+  end
+  ##
```
neugierige, line 122: 2 issues here: 1) `Lion` needs `include Carnivorous` in order to have access to those methods, 2) you are missing another `end` to indicate the end of the class

lib/diagnostic.rb
```diff
@@ -100,15 +115,18 @@ def eat_meat(food)
...
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# They can be very specific to the already specific instance of the Class
```
neugierige, line 129: In direct inheritance, you're limited to inheriting from a single source. In composition, you can draw from as many modules as you like.

neugierige:
0/6. Hi Ryan, you came so close on a few of the questions. Please look this over again and try to figure out the last question. Hint: you might want to review the "ruby-objects-self" lesson again. Please reach out if you want a 1:1.

PR #151, opened 2016-11-14
Attempt ruby-objects diagnostic

kiraamaa:
Use class resources to complete

Comfort: 4
Clarity: 5

I used my class notes to help with this assessment. 
Not sure how to fix/edit last problem.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -24,7 +24,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -100,20 +118,26 @@ def eat_meat(food)
...
-# your response as a comment here
+# Mixins allow you to draw from multiple sources/modules whereas direct inheritance
+# does not.
 ##
 
 ##
 # Fix the 'ComboAttack' class below so that calling
-# `ComboAttack.new.punch.move('left').kick.damage` doesn't cause an error.
+# `ComboAttack.new.punch.move('left').kick.damage` doesn't cause an error. <-- not sure how to do this?
 #
```
BenGitsCode, line 141: That's okay, it's a tough question and the wording could be reworked. 

``` ruby
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.get_possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
      1.5
    when ['kick', 'punch', 'up']
      2
    else
      1
    end
  end
end

```

Does this solution make sense to you?

BenGitsCode (commented):
5/6 good work!

PR #156, opened 2016-11-14
ruby objects dianostic complete

akadane87:


lib/diagnostic.rb
```diff
@@ -33,8 +33,19 @@
...
+  attr_read :location
+def initialize(name, age, location)
+  @name = "Dave"
+  @age = 32
+  @location = "Ohio"
+end
+end
+dave = Person.new('Dave')
 Response.dave = dave
+
```
BenGitsCode, line 48: Close!

``` ruby
class Person
  attr_reader :name, :age
  attr_writer :name, :location
  ## Alternatively,
  # attr_accessor :name
  # attr_reader :age
  # attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end
```

BenGitsCode (commented):
4/6 good work!

PR #157, opened 2016-11-14
Add response to diagnostic.

coffeeisamazing:
Add response to diagnostic.

Comfort: 3
Clarity: 2

I used Google and my class notes to help with this assessment.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -24,7 +24,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -142,7 +167,6 @@ def kick
     @damage *= multiplier
   end
 
-  private
   def multiplier
     case (moves)
     when ['punch', 'move left', 'kick']
```
BenGitsCode, line 172: 1/2 right!

``` ruby
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.get_possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
      1.5
    when ['kick', 'punch', 'up']
      2
    else
      1
    end
  end
end
```

does this make sense?

BenGitsCode (commented):
5/6 good work!

PR #159, opened 2016-11-14
Ruby Objects Diagnostic on 11/14/16

Sphilippe7:
comfort: 2
clarity: 4

I felt confident in what I 'needed' to do but actually doing it was
another situation. I'm getting there but I still need more
practice.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -33,7 +33,7 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = dave
 Response.dave = dave
 ##
 
@@ -43,7 +43,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def initialize
+    @solve_problems = "think think think"
+  end
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -33,7 +33,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = dave
 Response.dave = dave
 ##
```
BenGitsCode, line 38:
``` ruby
dave.location = 'Somerville'
Response.dave = dave
```

does this make sense?

lib/diagnostic.rb
```diff
@@ -145,7 +156,7 @@ def kick
   private
   def multiplier
     case (moves)
-    when ['punch', 'move left', 'kick']
+    when ['punch', 'left', 'kick']
       1.5
     when ['kick', 'punch', 'up']
       2
```
BenGitsCode, line 162:
``` ruby
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.get_possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
      1.5
    when ['kick', 'punch', 'up']
      2
    else
      1
    end
  end
end
```

does this solution make sense?

BenGitsCode (commented):
4/6

good work!

PR #162, opened 2016-11-14
Answer diagnostic questions

akradin:
4 on comfort 3 on clarity (except 2 on comfort on last one)
Used Ruby Docs

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -24,7 +24,15 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name
+  attr_writer :location
+  def initialize (name, location, age)
+    @name = name
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -33,8 +41,9 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new('Dave', 'Ohio', 32)
+dave.location = 'Somerville'
+Response.dave = <Person:0x007f9e5ee3d560 @location="Boston", @name="Dave">
 ##
```
BenGitsCode, line 47:
``` ruby
dave.location = 'Somerville'
Response.dave = dave
```

lib/diagnostic.rb
```diff
@@ -142,6 +168,10 @@ def kick
     @damage *= multiplier
   end
 
+  def get_possible_moves
+    p 'kick, move, punch'
+  end
+
```
BenGitsCode, line 174:
``` ruby
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.get_possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
      1.5
    when ['kick', 'punch', 'up']
      2
    else
      1
    end
  end
end
```

halfway there!

BenGitsCode (commented):
5/6

good work!

PR #163, opened 2016-11-14
Complete diagnostic

sujkid:
Done.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -24,7 +24,15 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -142,6 +167,10 @@ def kick
     @damage *= multiplier
   end
 
+  def self.get_possible_moves
+    "kick, move, punch"
+  end
+
   private
```
BenGitsCode, line 174: 1/2 right!

```
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.get_possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
      1.5
    when ['kick', 'punch', 'up']
      2
    else
      1
    end
  end
end
```

does this solution make sense?

BenGitsCode (commented):
5/6 good work!

PR #172, opened 2017-02-06
Respond to diagnostic

samordonez:
- Comfort: 4
- Clarity: 5

- I used google to reference ruby method chaining as a double check

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -136,25 +163,32 @@ def punch
...
+  def self.get_possible_moves
+    'kick, move, punch'
   end
 
   private
 
   def multiplier
     case moves
-    when `%w`['punch', 'move left', 'kick']
+    when `%w{punch, move left, kick}` # ['punch', 'move left', 'kick']
```
bernardlee, line 189: The grave accents will send this into the shell as a command.  Try inserting
```ruby
puts `ls`
```
into a Ruby file and running it.

bernardlee:
7/8

PR #174, opened 2017-02-06
Response

laurpaik-zz:


lib/diagnostic.rb
```diff
@@ -78,10 +93,12 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
-# your response as a comment here
```
bernardlee, line 84: `say_hello` is not defined on `HouseCat` or `Cat`, but it is defined on `Animal`, so it needs be looked up in its superclasses.

lib/diagnostic.rb
```diff
@@ -125,25 +151,25 @@ def eat_meat(food)
 
 # ComboAttack Class definition
 class ComboAttack
-  attr_reader :moves, :damage
+  attr_accessor :moves, :damage
```
bernardlee, line 154: This is unnecessary.

lib/diagnostic.rb
```diff
@@ -125,25 +151,25 @@ def eat_meat(food)
...
 
   def initialize
     @moves = []
     @damage = 0
   end
 
-  def punch
-    @moves << 'punch'
+  def self.punch
```
bernardlee, line 161: None of the existing methods need to be modified into class methods.  All of the public instance methods just need to return `self` to make the methods chainable.

bernardlee:
6/8

PR #175, opened 2017-02-06
Diagnostic complete

lbreguet:
-comfort: 3
-clarity: 4
-use of class notes and ruby doc

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_writer :location
+  attr_accessor :name
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +44,7 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +44,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
```
bernardlee, line 47: `dave`'s location property should be set to `'Somerville'` afterward.

lib/diagnostic.rb
```diff
@@ -78,10 +91,11 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat and I go 'meow'"
```
bernardlee, line 94: Missing comma after the class name.

lib/diagnostic.rb
```diff
@@ -105,15 +119,20 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+"Don't know."
```
bernardlee, line 135: Composition allows a class to draw from multiple sources whereas inheritance only allows a class to draw from one source.

lib/diagnostic.rb
```diff
@@ -136,16 +155,23 @@ def punch
...
   end
 
   def kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
+    self
+  end
+
+  def get_possible_moves
```
bernardlee, line 173: This should be a class method.  `def self.get_possible_moves`.

bernardlee:
4/8

PR #176, opened 2017-02-06
got through as much as I could

novakw267:
I am not sure if its all correct, I got through what I could. Setting up the difference between inheritance and mixins is still a little confusing on the actual syntax/structure. I think however conceptually I understand the differences. 

clarity 4
comfort 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
 ##
 
 ##
@@ -35,7 +45,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("dave", "32", "Ohio")
+dave.update_attribute(:location, Somerville)
 Response.dave = dave
 ##
 
@@ -45,7 +56,12 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+require_relative
+class Developer < Person
+
+Def solve_problems
+  puts "think think think"
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +45,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("dave", "32", "Ohio")
+dave.update_attribute(:location, Somerville)
```
bernardlee, line 49: `update_attribute` is not a method on `Person` and `Somerville` isn't declared.

lib/diagnostic.rb
```diff
@@ -26,7 +26,17 @@
...
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
```
bernardlee, line 39: Missing `end`.

lib/diagnostic.rb
```diff
@@ -45,7 +56,12 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+require_relative
```
bernardlee, line 59: This is unnecessary.

lib/diagnostic.rb
```diff
@@ -45,7 +56,12 @@
...
+class Developer < Person
+
+Def solve_problems
```
bernardlee, line 62: `Def` should be `def`.

lib/diagnostic.rb
```diff
@@ -78,12 +94,16 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go meow"
```
bernardlee, line 97: `meow` should be `'meow'`.

lib/diagnostic.rb
```diff
@@ -105,15 +125,22 @@ def eat_meat(food)
 end
 
 ##
-# your response here
```
bernardlee, line 108: This class definition is broken since it needs a closing `end`.  There needs to be a `roar` method that prints `'ROAR!'`.  Currently, the constructor sets an instance variable, `@roar` to the string `'ROAR!'`.

```ruby
class Lion < Cat
  include Carnivorous

  def roar
    puts 'ROAR!'
  end
end
```

bernardlee:
2/8

PR #177, opened 2017-02-06
Answer questions related with Ruby

wilsonkaya:


lib/diagnostic.rb
```diff
@@ -78,10 +89,13 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go meow"
```
bernardlee, line 92: `meow` should be `'meow'`.

lib/diagnostic.rb
```diff
@@ -105,15 +119,21 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+class Lion < Cat
+  include Carnivorus
```
bernardlee, line 123: `Carnivorus` should be `Carnivorous`.

lib/diagnostic.rb
```diff
@@ -105,15 +119,21 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# Mixins are preventing us form repeating our selves that resulst with D.R.Y
```
bernardlee, line 135: Composition allows a class to draw from multiple sources whereas inheritance only allows a class to draw from one source.

bernardlee:
3/8

PR #179, opened 2017-02-06
Submit Diagnostic

jwaskow:
Comfort: 4
Clarity: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -45,7 +55,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts 'think think think'
```
bernardlee, line 60: This will print the correct string to the console, but will return `nil`.

lib/diagnostic.rb
```diff
@@ -105,15 +120,21 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+class Lion < Cat
+  include Carnivorous
+  def roar
+    p "ROAR!"
```
bernardlee, line 126: This has the same issue as `Developer#solve_problems`.

bernardlee:
5/8

PR #180, opened 2017-02-06
Diagnostics - Ruby Object

zchenbonterra:
comfort: 3
Clarity: 5

I used require readings and ruby documentation for this. However, I
don't how to chain class methods. Please advise!

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +44,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'Somerville')
```
bernardlee, line 47: The instructions were to instantiate a new `Person` with a particular set of properties and then modify them afterward.

lib/diagnostic.rb
```diff
@@ -78,10 +91,11 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
-# your response as a comment here
+# The first interpolation would translate over to HouseCat and
```
bernardlee, line 97: This doesn't quite answer the question since there is no mention of the method lookup chain.  `say_hello` is not defined on `HouseCat`, so Ruby will check the superclass, `Cat`, for it.  Since it's not defined on `Cat` either, it looks up `Cat`'s superclass, `Animal`, finds it, and calls it.

bernardlee:
5/8

PR #181, opened 2017-02-06
diagnostic

dsweetser:
-took a good crack at diagnostic

Comfort: 4
Clarity: 5

Used class notes.  Had trouble troubleshooting Ruby code, I've started spending some time on CodeWars working on Ruby Kata to get more comfortable working with and understanding the language.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +44,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'Ohio')
+dave.location = 'Somerville'
 Response.dave = dave
 ##
 
@@ -45,7 +55,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    return "think think think"
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -133,19 +159,23 @@ def initialize
...
+    self.moves << 'kick'
+    self.damage += 10
+    self.damage *= multiplier
+  end
+
+  def get_possible_moves
+    return "kick, move, punch"
   end
 
   private
```
BenGitsCode, line 181:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

lib/diagnostic.rb
```diff
@@ -78,12 +92,15 @@ def initialize
...
+Response.housecat_noise = "I am a HouseCat, and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+# When we call say_hello the lookup chain goes from HouseCat to Cat to Animal
+# before it finds a response.  It then takes that command back to HouseCat,
+# which defines both self (HouseCat) and sound (meow), which it fills in and puts.
+# (added #'s because I was annoyed by atom coloring my response when I used words that are code')
```
bernardlee, line 103: This is fine since you should be running tests.

lib/diagnostic.rb
```diff
@@ -105,15 +122,24 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# A lot of this seems semantic to me: while you can get the same behavior
```
bernardlee, line 138: The primary advantage is that classes can include any number of modules but can only inherit from one class.

lib/diagnostic.rb
```diff
@@ -133,19 +159,23 @@ def initialize
   end
 
   def punch
-    @moves << 'punch'
-    @damage += 5
-    @damage *= multiplier
+    self.moves << 'punch'
```
bernardlee, line 162: This is unnecessary unless you specifically want to use the `self.variable` or `self.variable=` methods.

lib/diagnostic.rb
```diff
@@ -133,19 +159,23 @@ def initialize
   end
 
   def punch
```
bernardlee, line 161: `punch`, `move`, and `kick` should return `self` to make the methods chainable.

bernardlee:
6/8

PR #182, opened 2017-02-06
ruby object diagnostic

ggabrakadabra:
-answered questions

Comfort: 3
Clarity: 3

I think I'm getting better at Ruby but I couldn't figure out the last question completely. I think I did alright on the other questions though.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +45,9 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
+dave.location="Somerville"
+
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -45,7 +57,16 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+
+  def initialize
```
bernardlee, line 62: This constructor is unnecessary.

lib/diagnostic.rb
```diff
@@ -45,7 +57,16 @@
...
+    @solve_problems = solve_problems
+  end
+
+  def solve_problems
+    puts "think think think"
```
bernardlee, line 67: This will return `nil`, rather than the string, `'think think think'`, since `puts` only outputs to the console but doesn't have a return value.

lib/diagnostic.rb
```diff
@@ -78,12 +99,13 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = I am a HouseCat, and I go 'meow'
```
bernardlee, line 102: This code will be interpreted literally and cause an error since the string isn't wrapped in quotation marks.

lib/diagnostic.rb
```diff
@@ -78,12 +99,13 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+# This creates a new HouseCat, since it inherits from Cat which inherits from Animal,
```
bernardlee, line 107: Correct, but we want a mention of why `say_hello` works due to method lookups on superclasses.

lib/diagnostic.rb
```diff
@@ -126,6 +155,7 @@ def eat_meat(food)
 # ComboAttack Class definition
 class ComboAttack
   attr_reader :moves, :damage
+  attr_writer :moves, :damage
```
bernardlee, line 158: This will make `@moves` and `@damage` writable, but not readable.

bernardlee:
4/8

PR #183, opened 2017-02-06
Maura's Ruby Objects Diagnostic Response

ladypinetree:
Comfort: 3
Clarity:4

Used my notes and our trainings from last week.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :name :location
+  attr_accessor :age
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,8 +44,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new("Dave", 32, "Ohio")
+:location.dave = Somerville
 ##
 
 ##
@@ -45,7 +54,12 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+
+class Developer < Person
+    def solve_problems
+      puts 'think think think'
+    end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,16 @@
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :name :location
```
bernardlee, line 30: These symbols need to be separated with a comma.

bernardlee, line 30: `name` should be both readable and writable, so it should be declared with `attr_accessor`.  `location` should only be writable, so it should be declared with `attr_writable`.

lib/diagnostic.rb
```diff
@@ -26,7 +26,16 @@
...
+  attr_accessor :age
```
bernardlee, line 31: `age` should only be readable, so it should be declared with `attr_reader`.

lib/diagnostic.rb
```diff
@@ -35,8 +44,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new("Dave", 32, "Ohio")
+:location.dave = Somerville
```
bernardlee, line 48: This will attempt to invoke a method (or set a property) named `dave` on the symbol `:location` and set it equal to an identifier named `Somerville`.  It will cause an error.  `dave.location = 'Somerville'`.

lib/diagnostic.rb
```diff
@@ -45,7 +54,12 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+
+class Developer < Person
+    def solve_problems
+      puts 'think think think'
```
bernardlee, line 60: This will return `nil` since `puts` doesn't have a return value.  Since we want to return a string, we can explicitly return it with `return 'think think think'` or implicitly return it with `'think think think'`.

lib/diagnostic.rb
```diff
@@ -78,11 +92,14 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCar, and I go 'meow'"
```
bernardlee, line 95: "HouseCar" should be "HouseCat".

lib/diagnostic.rb
```diff
@@ -105,16 +122,23 @@ def eat_meat(food)
...
 
 ##
-# your response here
+class Lion < Cat
+  include Carnivorous
+  def roar
+    puts 'ROAR!'
+  end
+end
+end
```
bernardlee, line 131: This extraneous `end` will cause an error.

lib/diagnostic.rb
```diff
@@ -105,16 +122,23 @@ def eat_meat(food)
...
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
-##
+# # Using mixins allows you to include some methods when creating a new class that is similar or behaves in the
```
bernardlee, line 139: This is true, but it doesn't explain why composition is advantageous over inheritance.  The primary reason is that a class can only inherit from one other class.  Using composition, any number of modules can be included in a class.

bernardlee:
1/8

PR #184, opened 2017-02-06
ruby object diagnsotic response

cshea93:
Comfort : 4

Clarity : 4

I used google and class labs from last week.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,19 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+
+  end
+end
 ##
 
 ##
@@ -35,7 +47,7 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new ("Dave", 32 ,"Ohio")
 Response.dave = dave
 ##
 
@@ -45,7 +57,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts "think think think"
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +47,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new ("Dave", 32 ,"Ohio")
```
bernardlee, line 50: `dave`'s `location` should be set to `'Somerville'` afterward.

```ruby
dave.location = 'Somerville'
```

lib/diagnostic.rb
```diff
@@ -78,11 +94,12 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a Cat, and I go meow"
```
bernardlee, line 97: `meow` needs to be wrapped in single quotation marks.

lib/diagnostic.rb
```diff
@@ -78,11 +94,12 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
-##
+## because house_cat inherits the say_hello method from the animal class.
```
bernardlee, line 101: Be as specific as possible.  house_cat/House_cat and animal should be `HouseCat` and `Animal`, respectively.

lib/diagnostic.rb
```diff
@@ -105,15 +122,21 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# Using mixins/composition allows developers to keep their
```
bernardlee, line 138: The primary advantage is that classes can only inherit from one other class, while classes can include any number of modules.

lib/diagnostic.rb
```diff
@@ -45,7 +57,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts "think think think"
```
bernardlee, line 62: This will return `nil` since `puts` doesn't return a value.  You can use an explicit return with `return 'think think think'` or an implicit return with `'think think think'`.

bernardlee:
3/8

PR #185, opened 2017-02-06
Finish Diagnostic

newayliu1:
Comfort: 3
Clarity: 3

I used Google and my class notes to help with this diagnostic.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -105,15 +122,20 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# Mixins stores methods, is opening to all the classes.
```
bernardlee, line 138: Classes can include any number of modules, but classes can inherit from only one other class.

lib/diagnostic.rb
```diff
@@ -136,16 +158,23 @@ def punch
...
   end
 
   def kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
+    self
+  end
+
+  def get_possible_moves
```
bernardlee, line 176: This should be a class method.

```ruby
def self.get_possible_moves
  'kick, move, punch'
end
```

bernardlee:
6/8

PR #186, opened 2017-02-06
ruby objects diagnostic 2.6.2017

K0iram:
Comfort - 3
Clarity - 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name, :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
+
 ##
 
 ##
@@ -35,8 +45,10 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, Ohio)
+
 Response.dave = dave
+
 ##
 
 ##
@@ -45,7 +57,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts 'think think think'
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,17 @@
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name, :location
```
bernardlee, line 31: `location` should be write-only.

lib/diagnostic.rb
```diff
@@ -35,8 +45,10 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, Ohio)
+
```
bernardlee, line 49: `dave`'s `location` should be set to `Somerville` after it's instantiated.

```ruby
dave.location = 'Somerville'
```

lib/diagnostic.rb
```diff
@@ -45,7 +57,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts 'think think think'
```
bernardlee, line 62: This will return `nil` since `puts` doesn't return anything.  You can use an explicit return with `return 'think think 'think'` or an implicit return with `'think think think'`.

lib/diagnostic.rb
```diff
@@ -78,12 +94,12 @@ def initialize
...
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+The self in the say_hello method refers to the class name which is "HouseCat" and the sound looks into housecat to see the noise so it gets replaced with 'meow'
```
bernardlee, line 102: The answer that we're looking for involves the method lookup chain.  `say_hello` isn't defined on `HouseCat`, but since `HouseCat`'s superclass is `Cat`, Ruby checks for the existence of `say_hello` on `Cat`.  Since `say_hello` doesn't exist on `Cat`, it looks for it in its superclass, `Animal`, where it's defined.

lib/diagnostic.rb
```diff
@@ -105,15 +121,21 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+class Lion < Cat
```
bernardlee, line 124: This class doesn't have an instance method named `roar` defined.

lib/diagnostic.rb
```diff
@@ -105,15 +121,21 @@ def eat_meat(food)
...
+  include Carnivorous
+end
+
+  def roar
```
bernardlee, line 128: This is a standalone function definition.

lib/diagnostic.rb
```diff
@@ -105,15 +121,21 @@ def eat_meat(food)
...
+  end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+You dont have to define new methods for new classes
```
bernardlee, line 138: Classes can include any number of modules.  Classes can only inherit from one other class.

bernardlee:
1/8

PR #187, opened 2017-02-06
Completed Diagnostic

spwisner:
Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +45,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
+dave.location = "Somerville"
 Response.dave = dave
 ##
 
@@ -45,7 +56,13 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+
+class Developer < Person
+  def solve_problems
+    "think think think"
+  end
+end
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -105,15 +124,21 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# Mixins are used when things "behave as" rather than "is a" part of.
```
bernardlee, line 141: Classes can include any number of modules, but classes can only inherit from one other classes.

lib/diagnostic.rb
```diff
@@ -125,7 +150,7 @@ def eat_meat(food)
 
 # ComboAttack Class definition
 class ComboAttack
-  attr_reader :moves, :damage
+  attr_accessor :moves, :damage
```
bernardlee, line 153: This doesn't need to be changed.

lib/diagnostic.rb
```diff
@@ -148,8 +173,6 @@ def kick
     @damage *= multiplier
   end
 
-  private
```
bernardlee, line 151: This doesn't need to be changed.

lib/diagnostic.rb
```diff
@@ -148,8 +173,6 @@ def kick
...
-
   def multiplier
     case moves
     when `%w`['punch', 'move left', 'kick']
```
bernardlee, line 178: The grave accents should be removed on this line and the next line.

lib/diagnostic.rb
```diff
@@ -125,7 +150,7 @@ def eat_meat(food)
 
 # ComboAttack Class definition
 class ComboAttack
```
bernardlee, line 152:
```ruby
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.get_possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

bernardlee:
6/8

PR #188, opened 2017-02-06
complete diagnostic

SquirtleSquad1988:
Comfort: 4
Clarity: 4

Used class notes

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -45,7 +55,17 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def initialize(name, age, location)
```
bernardlee, line 59: This constructor is unnecessary since it's identical to the superclass's constructor.

lib/diagnostic.rb
```diff
@@ -78,10 +98,12 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = 'I am Animal and I go "meow"'
```
bernardlee, line 101: This should be `"I am a HouseCat, and I go 'meow'"`.

lib/diagnostic.rb
```diff
@@ -78,10 +98,12 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
-# your response as a comment here
+'sound would look to the current class whereas self.class.name would look
```
bernardlee, line 104: `self.class.name` would be evaluated to `HouseCat.class.name`, which would be evaluated to `HouseCat.name`, which would be evaluated to `"HouseCat"`, since `say_hello` is a superclass method being called from `HouseCat`.

bernardlee:
4/8

PR #189, opened 2017-02-06
Change classes, mixins - ruby

skylarkJ:


lib/diagnostic.rb
```diff
@@ -26,26 +26,36 @@
 # be writable.
 
 ##
-# your response here
+class Person
```
bernardlee, line 29: This class doesn't have a constructor.

```ruby
class Person
  attr_accessor :name
  attr_reader :age
  attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end
```

lib/diagnostic.rb
```diff
@@ -26,26 +26,36 @@
...
+  attr_writer :name, :location
+end
 ##
 
 ##
 # Instatiate a new Person object using arguments "Dave", 32, and "Ohio".
 # Then, in the next line, change this Person's location property to
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
-
+def initialize (Dave,32, Ohio)
```
bernardlee, line 39: This is defining a function called `initialize` with an invalid parameter name (32) and will cause an error.

bernardlee, line 39: We want to instantiate an instance of `Person` with the three given initial values and then modify its `location`.

```ruby
dave.location = 'Somerville'
```

lib/diagnostic.rb
```diff
@@ -26,26 +26,36 @@
...
+end
 ##
 # Create another class called Developer that inherits from Person.
 # Give it a new instance method called 'solve_problems',
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def instance
```
bernardlee, line 55: This defines an instance method called `instance` that sets an instance variable called `@solve_problems` to the string `'think think think'`.

lib/diagnostic.rb
```diff
@@ -26,26 +26,36 @@
...
 ##
-
+end
 ##
 # Create another class called Developer that inherits from Person.
 # Give it a new instance method called 'solve_problems',
 # which returns the string "think think think".
 
 ##
-# your response here
```
bernardlee, line 48:
```ruby
class Developer < Person
  def solve_problems
    'think think think'
  end
end
```

lib/diagnostic.rb
```diff
@@ -82,12 +92,20 @@ def initialize
...
+HouseCat is refering to a Cat class, Cat class is refering to Animal - animal has the sound nil.
 
 ##
 # Define a new class, 'Lion', which (a) inherits from 'Cat',
 # (b) uses the 'Carnivorous' module below as a mixin, and
 # (c) adds a new method called `roar`, which prints out "ROAR!"
+class Lion < Cat
+ include Carnivorous
+
+   def road
```
bernardlee, line 104: This should be "roar".

lib/diagnostic.rb
```diff
@@ -82,12 +92,20 @@ def initialize
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
-##
+HouseCat is refering to a Cat class, Cat class is refering to Animal - animal has the sound nil.
```
bernardlee, line 95: `say_hello` doesn't exist on `HouseCat` so Ruby looks it up on the superclass, `Cat`, which doesn't have it.  It then looks it up on `Cat`'s superclass, `Animal`, which does have it, and calls it.

bernardlee:
0/8

PR #190, opened 2017-02-06
Diagnostic response

AshlenPrice:
Comfort 3
Clarity 2.5
Used notes and Object repo  and Object self

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+#
+class Person
+  attr_accessor :name
+  attr_writer :location
+  attr_reader :age
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +45,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', '32', 'Ohio')
+dave.location = 'Somerville'
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -45,7 +56,12 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Deverloper < Person
+  def solve_problems
+    puts 'think think think'
```
bernardlee, line 61: This will return `nil` since `puts` doesn't return a value.  You can use an explicit return with `return 'think think think'` or an implicit return with `'think think think'`.

lib/diagnostic.rb
```diff
@@ -78,10 +94,14 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat and I go meow"
```
bernardlee, line 97: Missing comma after "HouseCat" and missing single quotation marks around "meow".

lib/diagnostic.rb
```diff
@@ -78,10 +94,14 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
-# your response as a comment here
+
+# I'm not very comfortable with self yet to know exactly why but I think it has to do
+# with the fact that in ruby its going to look up at the say_hello method that HouseCat inherits from Cat
```
bernardlee, line 102: `say_hello` isn't inherited from `Cat`, but from `Animal`, which is a superclass of `Cat`, which is a superclass of `HouseCat`.

lib/diagnostic.rb
```diff
@@ -105,15 +125,21 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# composition allows you to get information from multiple modules.
```
bernardlee, line 142: Yes, but this doesn't tell us about the advantages that it has over inheritance.  One advantage over inheritance is that with inheritance, classes can only inherit from one other class.

bernardlee:
3/8

PR #191, opened 2017-02-06
ruby-objects-diagnostic

cob0910:
responses - I could not figure out the errors when doing bin/rake test

Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +45,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'Ohio')
+dave.location = 'Somerville'
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -45,7 +56,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  DEV = Person.new(solve_problems: 'think think think')
```
bernardlee, line 60: This will cause an error since `Person`'s constructor expects three arguments.

```ruby
class Developer < Person
  def solve_problems
    'think think think'
  end
end
```

lib/diagnostic.rb
```diff
@@ -78,17 +93,24 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = I am a HouseCat, and I go 'meow'
```
bernardlee, line 96: This will be interpreted literally and cause an error.  This should be wrapped in quotation marks.

lib/diagnostic.rb
```diff
@@ -78,17 +93,24 @@ def initialize
...
 # your response as a comment here
 ##
-
+# HouseCat points to Cat which points to Animal which prints the sentence which gets the HouseCat name and teh HouseCat sound
 ##
 # Define a new class, 'Lion', which (a) inherits from 'Cat',
 # (b) uses the 'Carnivorous' module below as a mixin, and
 # (c) adds a new method called `roar`, which prints out "ROAR!"
 
+class Lion < Cat
```
bernardlee, line 107: This class doesn't include the `Carnivorous` module nor does it have a `roar` instance method.

lib/diagnostic.rb
```diff
@@ -78,17 +93,24 @@ def initialize
...
+  def initialize
```
bernardlee, line 108: This constructor will attempt to assign an instance variable `@roar`, to a local variable or parameter named `roar`, which doesn't exist, and cause an error.

bernardlee:
4/8

PR #192, opened 2017-02-06
Submission for diagnostic

alexkarasik:
Comfort: 2
Clarity: 2
I need to work on my syntax and structure a lot more.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,13 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+
+# I need a bit more time. I am not sure about some of the syntax and did not have time to look it up
+class Person
+ attr_accessor: name
+ attr_reader: age
+ attr_writer: location
+end
 ##
 
 ##
@@ -36,7 +42,9 @@ Response = OpenStruct.new
 
 ## replace nil with your response, then continue your work on the next line
 dave = nil
-Response.dave = dave
+Response.dave = dave.create(:name => "Dave", :age => 32, location => "Ohio")
+dave.location => "Somerville"
+dave = Person.new
 ##
 
 ##
@@ -45,7 +53,7 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,13 @@
 # be writable.
 
 ##
-# your response here
+
+# I need a bit more time. I am not sure about some of the syntax and did not have time to look it up
+class Person
```
bernardlee, line 31: This class is missing a constructor.

```ruby
class Person
  attr_accessor :name
  attr_reader :age
  attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end
```

lib/diagnostic.rb
```diff
@@ -36,7 +42,9 @@
 
 ## replace nil with your response, then continue your work on the next line
 dave = nil
-Response.dave = dave
+Response.dave = dave.create(:name => "Dave", :age => 32, location => "Ohio")
```
bernardlee, line 45: `dave` as an identifier doesn't exist and this will cause an error.

```ruby
```

lib/diagnostic.rb
```diff
@@ -36,7 +42,9 @@
...
+dave.location => "Somerville"
```
bernardlee, line 46: This syntax is invalid.

```ruby
dave.location = 'Somerville'
```

lib/diagnostic.rb
```diff
@@ -78,12 +86,12 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a Animal and I go nil"
```
bernardlee, line 89: This should be `"I am a HouseCat, and I go 'meow'"`.

lib/diagnostic.rb
```diff
@@ -45,7 +53,7 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+
```
bernardlee, line 56:
```ruby
class Developer < Person
  def solve_problems
    'think think think'
  end
end
```

lib/diagnostic.rb
```diff
@@ -78,12 +86,12 @@ def initialize
...
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a Animal and I go nil"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+#House inherits from cat, which inherits from animal, so the inheritance comes directly from the superclass, which is not exactly what this reponse wants to be doing I think. A mxin would have been better.
```
bernardlee, line 94: The inheritance chain is correct, but you didn't explain why the output is what it is.  `say_hello` doesn't exist on `HouseCat` or `Cat`, but does on `Animal`.  Since `say_hello` was found through the method lookup chain, it executes in the context of `HouseCat`, which is why `self.class.name` evaluates to `"HouseCat"`.

bernardlee:
2/8

PR #194, opened 2017-02-06
response

doremaxime:
comfort: 3
clarity: 4

Did not use any material.
Did not have time to answer last question but I did have a look at it.
In the repo it says to use grunt nag and grunt test, I thought that was for JS no?

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name
+  attr_writer :location
+
+  def instantiate(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +45,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', '32', 'Ohio')
+dave[2] = 'Somerville'
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +45,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', '32', 'Ohio')
+dave[2] = 'Somerville'
```
bernardlee, line 49: `dave` doesn't have `[]=` defined on it so this will cause an error.

```ruby
dave.location = 'Somerville'
```

lib/diagnostic.rb
```diff
@@ -26,7 +26,17 @@
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name
+  attr_writer :location
+
+  def instantiate(name, age, location)
```
bernardlee, line 34: The constructor method name is `initialize`.

lib/diagnostic.rb
```diff
@@ -78,11 +93,11 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I'm a HouseCat and I got 'meow'"
```
bernardlee, line 96: This should be `"I am a HouseCat, and I go 'meow'"`.

lib/diagnostic.rb
```diff
@@ -105,15 +120,24 @@ def eat_meat(food)
...
+end
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# mixins are more of behaviors, whereas if something has direct inheritance,
```
bernardlee, line 137: The primary reason is that classes can have many modules included, but only inherit from a single class.

bernardlee:
3/8

PR #195, opened 2017-02-06
diagnostis response.

moshiko1988:
Comfort: 4
Clarity: 4
i didnt have enough time to finish everything

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,8 +45,9 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new("Dave", 32, "Ohio")
+dave.location = "Somerville"
+Response.dave = dave.new
 ##
 
 ##
@@ -45,7 +56,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts "think think think"
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,8 +45,9 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new("Dave", 32, "Ohio")
+dave.location = "Somerville"
+Response.dave = dave.new
```
bernardlee, line 50: This will cause an error since `dave` is not a class and doesn't have a constructor.

lib/diagnostic.rb
```diff
@@ -78,10 +93,10 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go meow"
```
bernardlee, line 96: `meow` should be `'meow'`.

lib/diagnostic.rb
```diff
@@ -78,10 +93,10 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
-# your response as a comment here
+# HouseCat inherits from cat that inherits from Animal - #self calls the Instatiate method of HouseCat and @sound calls the sound of the HouseCat
```
bernardlee, line 99: `self` refers to the caller and doesn't call the constructor.  `@sound` is an instance variable on the constructors of `Animal` and `HouseCat`.

lib/diagnostic.rb
```diff
@@ -45,7 +56,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    puts "think think think"
```
bernardlee, line 61: This will return `'nil` since `puts` doesn't return a value.  You can use an explicit return with `return 'think think think'` or an implicit return with `'think think think'`.

lib/diagnostic.rb
```diff
@@ -105,7 +120,10 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+class Lion < Cat
+include Carnivorous
+@sound = "roar"
```
bernardlee, line 125: This sets an instance variable, `@sound`, on `Lion` with the value `"roar"`.

```ruby
class Lion < Cat
  include Carnivorous

  def roar
    puts 'ROAR!'
  end
end
```

bernardlee:
1/8

PR #196, opened 2017-02-06
response to diagnostic

Soboscott:
clarity 3
comfort 2
used  my notes, other git studies and learn ruby the hard way

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -28,6 +28,16 @@ Response = OpenStruct.new
 ##
 # your response here
 ##
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+def initialize
+  @name = name
+  @age = age
+  @location = locatin
+end
+end
 
 ##
 # Instatiate a new Person object using arguments "Dave", 32, and "Ohio".
@@ -35,8 +45,10 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new ('Dave', '32', 'Ohio')
+
+
+Response.dave = { 'Dave => Ohio'}
 ##
 
 ##
@@ -45,7 +57,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+def initialize(solve_problems)
+    @solve_problems = "think think think"
+  end
+end
 ##
 
 ##
@@ -78,12 +94,14 @@ end
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise =
+
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+NameError: uninitialized constant Response
+because "noise" was never initialized
 ##
 # Define a new class, 'Lion', which (a) inherits from 'Cat',
 # (b) uses the 'Carnivorous' module below as a mixin, and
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -28,15 +28,27 @@
...
 # your response here
 ##
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+def initialize
+  @name = name
+  @age = age
+  @location = locatin
```
bernardlee, line 38: This should be `@location = location`.

lib/diagnostic.rb
```diff
@@ -28,15 +28,27 @@
...
 # Then, in the next line, change this Person's location property to
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new ('Dave', '32', 'Ohio')
+
+
+Response.dave = { 'Dave => Ohio'}
```
bernardlee, line 51: This will cause a syntax error since it's being interpreted as a hash literal with the key `'Dave => Ohio'` with no `:` or `=>` to indicate the value.  We want to modify `dave`'s `location` and set it to `'Somerville'`.

```ruby
dave.location = 'Somerville'
Response.dave = dave
```

lib/diagnostic.rb
```diff
@@ -78,12 +94,14 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise =
```
bernardlee, line 97: `"I am a HouseCat, and I go 'meow'"`

lib/diagnostic.rb
```diff
@@ -45,7 +57,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+def initialize(solve_problems)
```
bernardlee, line 61: This creates a constructor that accepts one argument and sets an instance variable, `@solve_problems`, to it.  What we want is an instance method named `solved_problems`.

```ruby
class Developer < Person
  def solve_problems
    'think think think'
  end
end
```

lib/diagnostic.rb
```diff
@@ -105,15 +123,25 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+class Lion < Cat
+include Carnivorous
+def initialize (roar)
```
bernardlee, line 128: This is a constructor.  We want an instance method named `roar` that prints `ROAR!` to the console.

```ruby
class Lion < Cat
  include Carnivorous

  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -105,15 +123,25 @@ def eat_meat(food)
...
+
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+inheritance is used for "is-a " situations and mixins
```
bernardlee, line 140: The primary advantage is that classes can included any number of modules, but can only inherit from a single class.

lib/diagnostic.rb
```diff
@@ -78,12 +94,14 @@ def initialize
...
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise =
+
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+NameError: uninitialized constant Response
```
bernardlee, line 103: The output should be `"I am a HouseCat, and I go 'meow'"`.  Since `say_hello` isn't defined on `HouseCat`, Ruby tries to find it in its superclass, `Cat`.  It doesn't exist on `Cat` either so it checks its superclass, `Animal`.  `say_hello` exists on `Animal` so it's executed with `HouseCat` as the context.  `@sound` is an instance variable on `HouseCat`.

bernardlee:
0/8

PR #197, opened 2017-04-18
Add responses to ruby objects diagnostic

jim-moody:
Comfort: 5
Clarity: 5

I used `pry` for the HouseCat question and then just used the tests to check my work for the rest of the questions.  It looks like there was a separate issue with the ComboAttack class that wasn't part of the questions which you can see on lines 188, 190 of the PR

Also - get_possible_moves causes a linter warning :)

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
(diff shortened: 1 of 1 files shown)


Jcornmanhomonoff:
8/8! SUPER AWESOME JOB!!!! 🦄

PR #198, opened 2017-04-18
Completes lib/diagnostic.rb

buffalo0707:
Clarity: 5
Comfort: 4

Feeling good with the concepts of Objects in Ruby, but I am not sure how
they will be used in a real application. I guess I'll find out during
the next project...

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -78,7 +93,7 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
```
Jcornmanhomonoff, line 99: Ruby first looks for the `#say_hello` on the instance, and doesn't find it there or on HouseCat. The next place to look is Cat, since HouseCat inherits from Cat. Ruby doesn't find `#say_hello` there either, so it then looks at Animal, where it finds the method. When it executes the method, it is still executed on the instance of HouseCat, and when we created the instance, `@sound` was set to `'meow'`.

Jcornmanhomonoff:
7/8! Fantastic work! 🦄

PR #199, opened 2017-04-18
Commit for Ruby Objects Diagnostic

deeforte:
commit for April 18th diagnostic on ruby objects.
comfort: 4
clarity: 4
used the exercises from the ruby objects training rb files

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_accessor :name
+  attr_writer :location
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +44,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'Ohio')
+location.dave = 'Somerville'
 Response.dave = dave
 ##
 
@@ -45,7 +55,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    return 'think think think'
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +44,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'Ohio')
+location.dave = 'Somerville'
```
Jcornmanhomonoff, line 48: `dave.location = 'Somerville'`

lib/diagnostic.rb
```diff
@@ -102,18 +116,24 @@ def eat_meat(food)
...
-end
 
 ##
-# your response here
+class Lion
+  include Carnivorous
+  def roar
+    p 'ROAR!'
+  end
+end
```
Jcornmanhomonoff, line 126:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -102,18 +116,24 @@ def eat_meat(food)
...
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+Mixins are used when the class including it shares only a minor attribute.
+For example people and computer both sleep but otherwise they share no other
+simlarities.
```
Jcornmanhomonoff, line 136: In direct inheritance, you're limited to inheriting from a single source. In composition, you can draw from as many modules as you like.

lib/diagnostic.rb
```diff
@@ -160,4 +180,7 @@ def multiplier
       1
     end
   end
+  def get_possible_moves
+    return "kick, move, punch"
+  end
 end
```
Jcornmanhomonoff, line 186:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
4/8! Not bad! 🦄

PR #200, opened 2017-04-18
Completed diagnostic

Ryaong792:
did not finish, had to look up information to confirm some
answers.  Fist of 5 would still be a 3.  Just slow in answering.
did not get to see the last few questions due to time.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+attr_accessor :name
+attr_reader :age
+attr_writer :location
+
+def initialize(name, age, location)
+  @name = name
+  @age = age
+  @location = location
+end
+end
 ##
 
 ##
@@ -35,7 +45,10 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'ohio')
+dave.locaton = 'Somervile'
+
+def dave
 Response.dave = dave
 ##
 
@@ -45,7 +58,13 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+
+class Developer < Person
+  def solve_problems
+    p "think think think"
+  end
+end
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +45,10 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new('Dave', 32, 'ohio')
+dave.locaton = 'Somervile'
+
+def dave
```
Jcornmanhomonoff, line 51: don't need `def dave`

lib/diagnostic.rb
```diff
@@ -78,12 +97,15 @@ def initialize
...
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+the say_hello pulls from the class animal.  since the housecat is the chid
+of the cat is the chid of anial it will put the string of "I am a etc etc"
+{self.class.name} in the string will use the housecat and the sound will use
+the @sound for house cat = string for meow
 ##
```
Jcornmanhomonoff, line 109:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -78,12 +97,15 @@ def initialize
...
 # Define a new class, 'Lion', which (a) inherits from 'Cat',
```
Jcornmanhomonoff, line 110: In direct inheritance, you're limited to inheriting from a single source.

lib/diagnostic.rb
```diff
@@ -78,12 +97,15 @@ def initialize
...
 # (b) uses the 'Carnivorous' module below as a mixin, and
```
Jcornmanhomonoff, line 111:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
4/8! Not bad! 🦄

PR #202, opened 2017-04-18
Response

andyjzhong:
Clarity: 4/5
Comfort: 3/5

Used class notes.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,13 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,8 +41,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new('Dave', '32', 'Ohio')
+dave.location = 'Somerville'
 ##
 
 ##
@@ -46,6 +52,11 @@ Response.dave = dave
 
 ##
 # your response here
+class Developer < Person
+  def solve_problems
+    puts 'think think think'
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,13 @@
...
 
 ##
-# your response here
+class Person
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
```
Jcornmanhomonoff, line 35:
```rb
class Person
  attr_reader :name, :age
  attr_writer :name, :location
  ## Alternatively,
  # attr_accessor :name
  # attr_reader :age
  # attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end
```

lib/diagnostic.rb
```diff
@@ -35,8 +41,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new('Dave', '32', 'Ohio')
+dave.location = 'Somerville'
```
Jcornmanhomonoff, line 45: don't need to put 32 in quotes since it's not a string

lib/diagnostic.rb
```diff
@@ -106,6 +119,13 @@ def eat_meat(food)
 
 ##
 # your response here
+class Lion < Cat
+  def roar
+    puts "ROAR!"
+  end
+end
+
```
Jcornmanhomonoff, line 127:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -161,3 +182,5 @@ def multiplier
     end
   end
 end
+
+# This is a cool problem to solve -- I will need to revisit this one.
```
Jcornmanhomonoff, line 186:
```rb

class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end

```

lib/diagnostic.rb
```diff
@@ -161,3 +182,5 @@ def multiplier
     end
   end
 end
+
```
Jcornmanhomonoff, line 185: In direct inheritance, you're limited to inheriting from a single source.

Jcornmanhomonoff:
4/8! Not bad! 🦄

PR #204, opened 2017-04-18
study for april 18

awoodrum87:


lib/diagnostic.rb
```diff
@@ -26,7 +26,17 @@
...
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+end
+
```
Jcornmanhomonoff, line 39: Correct, but we usually initialize after determining writer/reader ability.

lib/diagnostic.rb
```diff
@@ -102,6 +119,10 @@ def eat_meat(food)
       puts 'Yuck!'
     end
   end
+
+class Lion < Cat
+  include Carnivorous
+
```
Jcornmanhomonoff, line 125:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -102,6 +119,10 @@ def eat_meat(food)
...
 end
```
Jcornmanhomonoff, line 127: In direct inheritance, you're limited to inheriting from a single source.

lib/diagnostic.rb
```diff
@@ -102,6 +119,10 @@ def eat_meat(food)
...
 
 ##
```
Jcornmanhomonoff, line 128:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
5/8! Good work! 🦄

PR #205, opened 2017-04-18
Response to Ruby Objects Diagnostic

rrdaniels85:
Comfort: 3/5
Clarity: 4/5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -105,13 +126,25 @@ def eat_meat(food)
...
 
+# It allows your code to inherit functionality from multiple modules rather than
+# just a parent class. With inheritance, I believe you can only inherit through
+# the single chain of class. However, with mixins, you can pull whatever
+# functionality you like into a class from multiple modules without having to
+# ensure the correct inheritance path. It jus gives you more flexibility.
+
 ##
 # your response as a comment here
 ##
```
Jcornmanhomonoff, line 150:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
7/8! Fantastic work! 🦄

PR #206, opened 2017-04-18
Ruby Object Diagnostic

cpm5041:
Clarity: 3
Comfort: 2

I felt ok during most of this diagnostic but would need more time
to carefully test and walk through each item, specifically the last one.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,8 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
-##
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer:location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+  ##
 
 ##
 # Instatiate a new Person object using arguments "Dave", 32, and "Ohio".
@@ -35,7 +43,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
+dave.location = "Somerville"
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,16 +26,25 @@
...
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer:location
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+  ##
```
Jcornmanhomonoff, line 38:
```rb
class Person
  attr_reader :name, :age
  attr_writer :name, :location
  ## Alternatively,
  # attr_accessor :name
  # attr_reader :age
  # attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end
```

lib/diagnostic.rb
```diff
@@ -45,8 +54,10 @@
 # which returns the string "think think think".
 
 ##
-# your response here
-##
+class Developer < Person
+  def solve_problems
+    puts "think think think"
+  end##
```
Jcornmanhomonoff, line 60: missing closing `end`

lib/diagnostic.rb
```diff
@@ -105,15 +116,25 @@ def eat_meat(food)
...
+Inheritance should only be used when:
+
+Both classes are in the same logical domain
+The subclass is a proper subtype of the superclass
+The superclass’s implementation is necessary or appropriate for the subclass
+The enhancements made by the subclass are primarily additive.
+^google helped me out on this one
 ##
 
 ##
```
Jcornmanhomonoff, line 140:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
5/8! Great work! 🦄

PR #207, opened 2017-04-18
Submitting REsponse

sjruby:
Need to spend more time on Ruby modules in the next workshop, I kind of
get teh idea but am still not comfortable coding it myself

Comfort 2/5
Clarity 3/5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :age
+  attr_writer :location
+  attr_accessor :name
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +44,7 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +44,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
```
Jcornmanhomonoff, line 47: Missing: 
`dave.location = 'Somerville'`

lib/diagnostic.rb
```diff
@@ -103,7 +127,12 @@ def eat_meat(food)
     end
   end
 end
+# I'm making a lion
+class Lion < Cat
+  include Carnivorous
+end
 
+Lion.new.roar
```
Jcornmanhomonoff, line 135:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -136,16 +166,19 @@ def punch
...
     @moves << "move #{direction}"
+    self
   end
 
   def kick
     @moves << 'kick'
     @damage += 10
     @damage *= multiplier
+    self
   end
```
Jcornmanhomonoff, line 183:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
5/8! Good job! 🦄

PR #208, opened 2017-04-18
Adds responses to diagnostic.rb for ruby objects

jkjellan:
Feeling more comfortable, but ran out of time.

Comfort:  3.5
Clarity: 5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,18 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -152,9 +176,9 @@ def kick
...
   def multiplier
     case moves
-    when `%w`['punch', 'move left', 'kick']
+    when %w(punch, move left, kick)
       1.5
-    when `%w`['kick', 'punch', 'up']
+    when %w(kick, punch, up)
       2
     else
       1
```
Jcornmanhomonoff, line 184:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
7/8! Great work! 🦄

PR #210, opened 2017-04-18
Ruby Morning Diagnotsitc answers on classes, self and inheritance.

glasneed:


lib/diagnostic.rb
```diff
@@ -26,7 +26,21 @@
...
+
+class Person
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
+end
+
```
Jcornmanhomonoff, line 43:
```rb
class Person
  attr_reader :name, :age
  attr_writer :name, :location
  ## Alternatively,
  # attr_accessor :name
  # attr_reader :age
  # attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end
```

lib/diagnostic.rb
```diff
@@ -105,15 +138,28 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+class Lion <  Cat
+  includes Carnivorous
+  def roar
+  returns "ROAR!"
+  end
+end
```
Jcornmanhomonoff, line 146: Singular `include`
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -123,6 +169,8 @@ def eat_meat(food)
 # Then, create a new class method called "get_possible_moves",
 # which returns the string "kick, move, punch"
 
+NO TIME LEFT TO DO THIS QUESTION!!
+
 # ComboAttack Class definition
 class ComboAttack
   attr_reader :moves, :damage
```
Jcornmanhomonoff, line 176:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
5/8! Good job! 🦄

PR #211, opened 2017-04-18
Ruby object diagnostic

antleo1995:
Answered the questions, only got tripped up by the last one. All the
rest was done from memory - very little google needed.

Calling it a 4 and a 4.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize (name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -78,12 +92,12 @@ def initialize
...
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat, and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+# Self.class.name refers to the class name of whatever is calling it.
```
Jcornmanhomonoff, line 100: Ruby first looks for the `#say_hello` on the instance, and doesn't find it there or on HouseCat. The next place to look is Cat, since HouseCat inherits from Cat. Ruby doesn't find `#say_hello` there either, so it then looks at Animal, where it finds the method. When it executes the method, it is still executed on the instance of HouseCat, and when we created the instance, `@sound` was set to `'meow'`.

Jcornmanhomonoff:
7/8! Great work! 🦄

PR #212, opened 2017-04-18
Add responses to ruby objects diagnostic

lisamsmith100:
FoF: 3.5, 3.5
I used solution branches of lessons and pry for responses.  need more
time.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,29 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_reader :name, :age
+  attr_writer :name, :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
+  def name
+    @name
+  end
+
+  def age
+    @age
+  end
+
+  def location=(location)
+    @surname = surname
+  end
+end
+
 ##
 
 ##
@@ -35,7 +57,7 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
 Response.dave = dave
 ##
 
@@ -45,7 +67,20 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def initialize(name, age, location)
+    super(name, age, location)
+  end
+
+  def solve_problems
+    puts string = 'think think think'
+  end
+end
+
+# define Class Ford
+class Ford
+end
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lisamsmith100:
I'd like to spend more time with mixins.

lib/diagnostic.rb
```diff
@@ -26,7 +26,29 @@
...
+
+  def age
+    @age
+  end
+
+  def location=(location)
+    @surname = surname
+  end
+end
+
```
Jcornmanhomonoff, line 51: Correct, but 
```rb
def name
    @name
  end

  def age
    @age
  end

  def location=(location)
    @surname = surname
  end
```
 is unnecessary.

lib/diagnostic.rb
```diff
@@ -35,7 +57,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave", 32, "Ohio")
```
Jcornmanhomonoff, line 60: `dave.location = 'Somerville'`

lib/diagnostic.rb
```diff
@@ -45,7 +67,20 @@
...
+
+  def solve_problems
+    puts string = 'think think think'
+  end
+end
+
+# define Class Ford
+class Ford
+end
+
```
Jcornmanhomonoff, line 83: You do not need to initialize again since Developer is inheriting from Person.

lib/diagnostic.rb
```diff
@@ -105,7 +144,10 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+require_relative 'Cat'
+require_relative 'Carnivorous'
+class Lion < Carnivorous
+  def
```
Jcornmanhomonoff, line 150:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end

```

lib/diagnostic.rb
```diff
@@ -105,7 +144,10 @@ def eat_meat(food)
...
 ##
 
 # #
```
Jcornmanhomonoff, line 153:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
4/8! Not bad! 🦄

PR #216, opened 2017-04-18
Responses: Ruby Objects Diagnostic

naributler:
Clarity: 3
Comfort: 2

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,11 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+end
 ##
 
 ##
@@ -35,7 +39,19 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(location)
+    @name = 'Dave'
+    @age = 32
+    @location = location
+  end
+end
+
+dave = Person.new('Somerville')
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,11 @@
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+end
```
Jcornmanhomonoff, line 33:
```rb
class Person
  attr_reader :name, :age
  attr_writer :name, :location
  ## Alternatively,
  # attr_accessor :name
  # attr_reader :age
  # attr_writer :location

    @name = name
    @age = age
    @location = location
  end
end

```

lib/diagnostic.rb
```diff
@@ -35,7 +39,19 @@
...
+  attr_writer :location
+
+  def initialize(location)
+    @name = 'Dave'
+    @age = 32
+    @location = location
+  end
+end
+
+dave = Person.new('Somerville')
```
Jcornmanhomonoff, line 54:
```rb
dave.location = 'Somerville'
Response.dave = dave

```

lib/diagnostic.rb
```diff
@@ -148,16 +180,20 @@ def kick
...
       2
     else
       1
     end
   end
+
+  def get_possible_moves
+    puts "kick, move, punch"
+  end
 end
```
Jcornmanhomonoff, line 199:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
5/8! Great work! 🦄

PR #217, opened 2017-04-18
Commit to the Ruby Objects Diagnostic

alzateja:
Comfort: 3
Clarity: 3

I ran out of time on the diagnostic. Still not 100% comfortable with
Ruby.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name ,age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
 ##
 
 ##
@@ -35,7 +44,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dave" ,32, "Ohio")
+dave.location = "Somerville"
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,16 @@
...
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name ,age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
```
Jcornmanhomonoff, line 38: missing `end`

lib/diagnostic.rb
```diff
@@ -105,15 +124,25 @@ def eat_meat(food)
...
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# You would use composition when you want you want to add a behavior that may
+# apply to very distinct classes.  This would be helpful in a case where you want
+# to add behavior where inheritance doesn't make sense like a person and cat eating.
```
Jcornmanhomonoff, line 145: In direct inheritance, you're limited to inheriting from a single source. In composition, you can draw from as many modules as you like.

lib/diagnostic.rb
```diff
@@ -105,15 +124,25 @@ def eat_meat(food)
...
 ##
 
 ##
```
Jcornmanhomonoff, line 148:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
6/8! Nice work! 🦄

PR #218, opened 2017-04-18
Fallon diagnostic

jonifallon:
Ran out of time

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,18 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer: location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+
+
 ##
 
 ##
@@ -35,8 +46,8 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new(name: "Dave", age: 32, location: "Ohio")
+Response.dave.location = "Somerville"
 ##
 
 ##
@@ -45,7 +56,13 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def initialize(name, age, location)
+  end
+  def solve_problems
+    puts "think think think"
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,18 @@
...
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer: location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
```
Jcornmanhomonoff, line 38: missing `end`

lib/diagnostic.rb
```diff
@@ -35,8 +46,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new(name: "Dave", age: 32, location: "Ohio")
+Response.dave.location = "Somerville"
```
Jcornmanhomonoff, line 50:
```rb
dave.location = 'Somerville'
Response.dave = dave
```

lib/diagnostic.rb
```diff
@@ -45,7 +56,13 @@
...
 
 ##
-# your response here
+class Developer < Person
+  def initialize(name, age, location)
+  end
+  def solve_problems
+    puts "think think think"
+  end
+end
```
Jcornmanhomonoff, line 65: Correct, however you don't need to initialize name, age, and location since it's already inheriting from Person & Person has those properties.

lib/diagnostic.rb
```diff
@@ -105,7 +123,14 @@ def eat_meat(food)
...
 ##
-# your response here
+class Lion < Cat
+  def initialize
+  end
+  #ADD MIXIN HERE - RAN OUT OF TIME
+  def roar
+    puts "ROAR!"
+  end
+end
```
Jcornmanhomonoff, line 133:
```rb
class Lion < Cat
  include Carnivorous
  def roar
    puts 'ROAR!'
  end
end
```

lib/diagnostic.rb
```diff
@@ -105,7 +123,14 @@ def eat_meat(food)
...
 ##
 
 # #
```
Jcornmanhomonoff, line 136:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
4/8. Not bad! If you have any questions please contact a consultant. 🦄

PR #219, opened 2017-04-18
Ruby Objects Diagnostic

brianedmunds:
Comfort: 3
Clarity: 3

I used my Google and my notes for this diagnostic.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person (name, age, location)
+  attr_accessor: name
+  attr_reader: age
+  attr_writer: location
+  def initialize (age, name location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
@@ -35,7 +44,9 @@ Response = OpenStruct.new
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new
+dave.age = 32
+dave.location = Somerville
 Response.dave = dave
 ##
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +44,9 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new
+dave.age = 32
+dave.location = Somerville
```
Jcornmanhomonoff, line 49:
```rb
dave.location = 'Somerville'
```

lib/diagnostic.rb
```diff
@@ -105,15 +122,21 @@ def eat_meat(food)
...
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+Mixins allow you to make chunks of code that are reusable across multiple classes.  This allow developers to keep their code DRY.
 ##
 
 ##
```
Jcornmanhomonoff, line 142:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
6/8! Fantastic job! 🦄

PR #220, opened 2017-04-18
Response to Diagnostic

danielleletarte:
2/2

It took me a long time to understand what the directions were asking and complete the first question. Will revisit the rest later in workshop.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -57,7 +57,11 @@ Response.dave = dave
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    'think think think'
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


danielleletarte:
I took another stab at this during lunch and, with ore time, was able to complete it. Still need a lot of practice working with Ruby.

lib/diagnostic.rb
```diff
@@ -125,39 +147,50 @@ def eat_meat(food)
...
-    when `%w`['kick', 'punch', 'up']
+    when %w('kick' 'punch' 'up')
       2
     else
       1
     end
   end
 end
+
+ComboAttack.get_possible_moves
```
Jcornmanhomonoff, line 196:
```rb
class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
7/8! Fantastic job! 🦄

PR #221, opened 2017-04-18
Response

donpowers:


lib/diagnostic.rb
```diff
@@ -35,8 +44,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new("Dave", 32, "Ohio")
+dave.location = "Somerville"
```
Jcornmanhomonoff, line 48: Missing assignment of modified Person
`Response.dave = dave`

lib/diagnostic.rb
```diff
@@ -105,15 +120,21 @@ def eat_meat(food)
 end
 
 ##
-# your response here
+Class Lion < Cat
+  include Carnivorous
+  def roar
+      'ROAR!'
```
Jcornmanhomonoff, line 126: `puts 'ROAR!'`

lib/diagnostic.rb
```diff
@@ -105,15 +120,21 @@ def eat_meat(food)
...
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+When the functionality can be used amongst several different classes, that
+naturally don't form a class.
 ##
 
 ##
```
Jcornmanhomonoff, line 140:
```rb

class ComboAttack
  attr_reader :moves, :damage

  def initialize
    @moves = []
    @damage = 0
  end

  def punch
    @moves << 'punch'
    @damage += 5
    @damage *= multiplier
    self
  end

  def move(direction)
    @moves << "move #{direction}"
    self
  end

  def kick
    @moves << 'kick'
    @damage += 10
    @damage *= multiplier
    self
  end

  def self.possible_moves
    'kick, move, punch'
  end

  private

  def multiplier
    case moves
    when %w(punch move left kick)
      1.5
    when %w(kick punch up)
      2
    else
      1
    end
  end
end
```

Jcornmanhomonoff:
5/8! Great job 🦄

PR #228, opened 2017-04-25
Add responses to Ruby Objects Diagnostic

nriser:
Diagnostic is currently incomplete. Still working on it.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader  :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -121,15 +121,22 @@ def eat_meat(food)
...
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# Using mixins is more popular than inheritance because you don’t have to think
+# about the relationship between classes.
 ##
```
scottyscripts, line 140: This is true! Mixins also allow us to inherit behavior from multiple modules, where as with inheritance, we can only inherit from one place (The Class we inherit from)

lib/diagnostic.rb
```diff
@@ -35,8 +45,8 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+dave = Person.new("Dave", 32, "Ohio")
+Response.dave = <Person:0x007f9e9302f538 @age=32, @location="Ohio", @name="Dave">
 ##
```
scottyscripts, line 50: how would we update the location properties value from outside the class?

scottyscripts:
7/8 Nice!

PR #230, opened 2017-04-25
WIP answers

TronB:


lib/diagnostic.rb
```diff
@@ -46,6 +57,10 @@
 
 ##
 # your response here
+Class Developer < Person
+def solve_problems
+  yield "think think think"
```
scottyscripts, line 62: Here, you don't need the `yield` keyword. If we wanted to return, we would need no keyword at all!

lib/diagnostic.rb
```diff
@@ -106,12 +128,22 @@ def eat_meat(food)
 
 ##
 # your response here
+class Lion < Cat
+  include Carnivorous
+end
+def roar
+  puts "ROAR!"
+end
 ##
```
scottyscripts, line 137: move that `end` on line 133 to the end. Make sure all methods and include are part of the Class!

scottyscripts:
7/8 Great job!

PR #231, opened 2017-04-25
Answers questions

Anddrew914:


lib/diagnostic.rb
```diff
@@ -105,15 +120,21 @@ def eat_meat(food)
...
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# Composition allows you to include functions that are simliar in objects that are
+# not neccessarily related.
 ##
```
scottyscripts, line 138: True! It also allows us to inherit behavior from multiple places instead of just from one as we do with inheritance!

scottyscripts:
7/8 Great job!

PR #234, opened 2017-04-25
add response in diagnostic.rb

aprilsourz:


lib/diagnostic.rb
```diff
@@ -45,7 +55,11 @@
 # which returns the string "think think think".
 
 ##
-# your response here
+class Developer < Person
+  def solve_problems
+    return "think think think"
```
scottyscripts, line 60: Nice! remember that we don't need the `return` keyword in Ruby
Ruby methods always try to return something, so `"think think think"` will do

scottyscripts:
7.5/8 Great job!! You are really close on the last one!

PR #235, opened 2017-04-25
diagnostic

eliottenos:
Comfort: 2
Clarity: 2

I used ruby docs, stackoverflow and tutorialspoint


scottyscripts:
I would definitely recommend requesting a one on one for this material. This stuff is super important to get down especially moving forward with Rails and into project 2 and we want to make sure you have this down!

PR #236, opened 2017-04-25
answers

prankmode:
comfort: 4
clarity: 5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -78,11 +93,16 @@ def initialize
 
 ## What will be the output from calling `HouseCat.new.say_hello`?
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = 'meow'
```
scottyscripts, line 97: That will be part of the output! 
Take a look at the `say_hello` method

lib/diagnostic.rb
```diff
@@ -105,15 +125,21 @@ def eat_meat(food)
...
 ##
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+# You can bring in behavior and not data.
 ##
```
scottyscripts, line 143: One can also bring in behavior with inheritance! The real power of mixins is being able to inherit that behavior from multiple sources, where as with inheritance we are limited to to inheriting from a single source.

scottyscripts:
5.5/8 Great job on this!

PR #237, opened 2017-04-25
Most questions answered

d00medman:
Having trouble w/penultimate question. Pulling anyways.


scottyscripts:
7/8 Great job! I think that you could get that last one if you tried it! Give it a try!

PR #238, opened 2017-04-25
Pull Request for ruby-objects-diagnostic

noobiwankenoobi:
Answered all questions except last one-- ran out of time
I am pretty confident that the answers I gave are either accurate or
close to being accurate.
I would not say that I'm confident utilizing this material from
scratch yet.

Comfort: 3
Clarity: 4

I used the documentation and previous studies to help with my answers

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,16 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+
+class Person
+  attr_reader :age
+  attr_writer :location
+  attr_accessor :name
+  def initialize(name, age, location)
+    @age = age
+    @location = location
+    @name = name
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +44,9 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new("Dav", 32, "Ohio")
+  @location = "Somerville"
+
```
scottyscripts, line 49: Great job creating a new instance of Person!
to set the property value outside of the class we would use `dave.location = "Somerville"`

scottyscripts:
6.5/8 Nice work! After some more repetition you'll be able to do it from scratch no problem!

PR #239, opened 2017-04-25
Completes Ruby Object Diagnostic

sstone72389:
Comfort: 3
Clarity: 4

I used Google and my class notes

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,19 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+end
+
+
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -78,12 +94,12 @@ def initialize
...
 # replace nil with your response
-Response.housecat_noise = nil
+Response.housecat_noise = "I am a HouseCat and I go 'meow'"
 
 ## Explain why this would be the output, based on the method lookup chain.
 # your response as a comment here
 ##
-
+Because @sound was overwritten by the new class House_cat.
 ##
```
scottyscripts, line 103: We are not really overwriting @sound because when we instantiate a new instance of HouseCat,
sound is defined. 
The question is more so asking how we have access to `say_hello` even though it's not defined on HouseCat

lib/diagnostic.rb
```diff
@@ -136,16 +159,23 @@ def punch
...
     @damage += 10
     @damage *= multiplier
+    self
+  end
+
+  def.get_possible_moves
+    'kick, move, punch'
   end
 
   private
```
scottyscripts, line 181: You're  so close to getting it to run! 
Just need to alter `get_possible_moves`

scottyscripts:
6.5/8 Great job! Nice work on the last one! You are soo close!

PR #240, opened 2017-04-25
Response

mlynnes:
comfort: 3
clarity: 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,10 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -26,7 +26,10 @@
 # be writable.
 
 ##
-# your response here
+class Person
+  def initialize(name, age, location)
+  end
+end
 ##
```
scottyscripts, line 34: .5
You're half way there! Remember what goes inside the initialize method?
Hint: `@name = name`

scottyscripts:
3.5/8 You show you know the material well based on the first few questions! I bet you can get the rest of them!

PR #242, opened 2017-04-25
Stuck on getting rid of error

rowlandhill:
3 of 5 fist of five

I removed the quotes in the build array shortcut %w, but still
am trying to find out there error.  Getting this message:

NoMethodError: undefined method `move' for 5:Fixnum
from (pry):42:in `__pry__'

Will continue to look for issue, but submitting everything else

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+
+  def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +45,7 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
+dave = Person.new(name: "Dave", age: 32, location: "Ohio")
 Response.dave = dave
 ##
```
scottyscripts, line 51: -.5 Remember to read the whole question!
How would we change the location property to "Somerville"?

scottyscripts:
6.5/8 Awesome! I think you can get the last one!

PR #243, opened 2017-04-25
Responses for Ruby Objects Study

kellye2010:
Comfort: 4
Clarity: 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,17 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+#
+class Person
+  attr_reader :name
+  attr_reader :age
+  attr_reader :location
+    def initialize(name, age, location)
+    @name = name
+    @age = age
+    @location = location
+  end
+  end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -35,7 +45,13 @@
...
-dave = nil
+
+# dave = Hash.new
+dave = Person.new
+# dave = Hash.new('')
+dave = Person.new('')
+dave = {name: "Dave", age: 32, location: "Ohio"}
+
 Response.dave = dave
 ##
```
scottyscripts, line 57: Remember that when instantiating a new instance of an object we pass arguments to the .new method
How would you change a property?

lib/diagnostic.rb
```diff
@@ -105,9 +126,17 @@ def eat_meat(food)
...
+end
 
+class Lion < Cat
+  def roar
+    puts 'ROAR!'
+  end
+end
+##
+#Not finished
 # #
```
scottyscripts, line 140: remember that we include modules with the include keyword

scottyscripts:
2.5/8 . If you have questions about this please reach out to myself or another consultant!

PR #244, opened 2017-04-25
 Responses to ruby-object diagnostic

JLisaW:


lib/diagnostic.rb
```diff
@@ -35,8 +48,9 @@
 # "Somerville". Finally, assign the modified Person to `Response.dave` below.
 
 ## replace nil with your response, then continue your work on the next line
-dave = nil
-Response.dave = dave
+def dave=(name: "Dave", age: 32, location: "Ohio")
+  dave.location = "Somerville"
+Response.dave = dave.location
 ##
```
scottyscripts, line 54: make sure you use the Class.new syntax when instantiating dave as a new person. 
Also, dave doesn't need the def before it.
So `dave = Person.new("Dave", 32, "Ohio")`
Great job changing the location! -.5

lib/diagnostic.rb
```diff
@@ -45,7 +59,14 @@
...
-# your response here
+require_relative 'Person'
+
+class Developer < Person
+  def solve_problems
+    puts "think think think"
+  end
+end
+
 ##
```
scottyscripts, line 70: Great! 
Pay close attention to the language in the question. To return the string we don't need `puts` there
That will actually print the string vs returning it.

lib/diagnostic.rb
```diff
@@ -105,7 +126,12 @@ def eat_meat(food)
 end
```
scottyscripts, line 126: the say_hello method isn't defined on `HouseCat` 
so Ruby looks at `Cat` because `HouseCat` inherits from `Cat`
Ruby can't find the method on `Cat` but it looks on `Animal` because `Cat` inherits from `Animal`
Inheritance is so great!

lib/diagnostic.rb
```diff
@@ -105,7 +126,12 @@ def eat_meat(food)
...
 
 ##
-# your response here
+class Lion < Cat
+include
+  def roar
+    puts "ROAR!"
+  end
+end
 ##
```
scottyscripts, line 135: Nice!
don't forget the module name after include!
So `include Carniverous`

scottyscripts:
4.5/8 Nice work!

PR #245, opened 2017-04-25
Diagnostic Response

Pouk91:
Comfort: 2
Clarity: 3

Used google

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -26,7 +26,15 @@ Response = OpenStruct.new
 # be writable.
 
 ##
-# your response here
+class Person
+  attr_accessor :name
+  attr_reader :age
+  attr_writer :location
+  def initialize(name, age, location)
+  @name = name
+  @age = age
+  @location = location
+end
 ##
 
 ##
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -47,7 +56,11 @@
 ##
 # your response here
 ##
-
+class Developer < Person
+end
+def solve_problems
+  puts "think think think"
+end
 ##
```
scottyscripts, line 64: Nice! make sure method is in class! I assume this is just a typo...

lib/diagnostic.rb
```diff
@@ -102,18 +117,26 @@ def eat_meat(food)
...
 
 # #
 # What are some of the advantages of using composition (i.e. mixins)
 # over using direct inheritance?
 
 ##
-# your response as a comment here
+Provides namespace and prevents name clashes
 ##
```
scottyscripts, line 140: The real benefit comes in the ability to draw from many different modules instead of inheriting from one source

scottyscripts:
6/8 Great job on this! If you have any questions that make you feel uncomfortable on this topic let us know!

